PR #5, opened 2020-02-14
Sensio Network Grant Application

woss:
# Grant Application Checklist


woss:
Thanks for the reply. 

> The first question that comes to my mind is: How do you prevent people from uploading photos from others? 

Uploading any photo and obtaining the provable claim are two different things. The scope of the SensioNetwork project, at least the part we are applying for, is to define and create the rules and validations which must be followed in order to satisfy the copyright/ownership claim. 


> Regarding parachain, I probably would simply call it a substrate based chain for now, since the exact (economic) terms of a parachain aren't clear at the moment. 

Should i change the the name to be `Substrate based chain` instead of `Parachain` ?

>  I'm also not sure we should support you running the nodes (milestone 3). I personally would be more interested in the code or interface you use in combination with the runtime modules (in case you open-source it anyway).

Oh, for that i am sorry, I never meant you to pay for that. 😄 I saw it as a `deliverable` at the end of the milestone, not as paid one. I will remove it.

> Maybe this could be useful for other projects in our ecosystem (e.g. supply chain projects, etc.)

Maybe. If I am right these rules could be even implemented as a Pallet. That is just a thought, since i am new to substrate and whole polkadot architecture. Time will tell will that be useful or not.

woss:
> This sounds like an interesting project. 
Yes it does. 😉

Noc2:
> Uploading any photo and obtaining the provable claim are two different things. The scope of the SensioNetwork project, at least the part we are applying for, is to define and create the rules and validations which must be followed in order to satisfy the copyright/ownership claim.

Makes sense. Thanks for updating the application. It would be nice if you could also specify the milestones/runtime modules a bit more. You don't have to write a full specification since this likely going to change anyway, but since you already seem to have something developed in nodejs (feel free to integrate a link to it), you can probably provide more details here. Also, unit tests would be nice for each module and as I said some form of an interface (could be polkadot.js based).

PR #9, opened 2020-03-24
WebAssembly Runtimes Fuzzing

pventuzelo:
# Grant Application Checklist

### DIFF
--- a/applications/wasm_runtimes_fuzzing.md
+++ b/applications/wasm_runtimes_fuzzing.md
@@ -21,7 +21,7 @@ This project aim to improve security and resilience of WebAssembly runtimes and 
 * **Legal Structure:** Patrick Ventuzelo, 113 grande rue, 91290 Arpajon, FRANCE
 * **Team's Experience:**
 
-Patrick is an Independent Security Researcher specialized in vulnerability research, fuzzing, reverse engineering and program analysis. He is teacher of two training respectively about “WebAssembly Security” and “Rust Security“.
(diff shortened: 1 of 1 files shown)


folsen:
I just had one point of clarification, this is work that will go into Octopus right? So Octopus will be the tool to test all these wasm execution engines (we tend to call it that to not confuse it with the runtime term used in Polkadot/Substrate which is a wasm blob).

I think this would be very valuable, particular for new development.

> 2.  | CLI tool | Command line tool allowing execution of wasm modules through all runtimes. |  

This would also be quite valuable for us in testing the polkadot runtime (the wasm blob) essentially running a given blob through all possible runtimes and ensuring the result is the same for all.

pventuzelo:
This tool (WebAssembly Runtimes Fuzzing i.e WARF in short) will be focused on testing/fuzzing wasm execution VM implementation. [Octopus](https://github.com/pventuzelo/octopus) is more focused on analysis of wasm module/blob.

This project will be independent of Octopus even if personally I'm planning to use WARF inside octopus to dynamically detect security issues inside wasm module and make in-process fuzzing of wasm module using the CLI tool.

Thx @folsen for the precision regarding polkadot runtime naming.

Noc2:
Hi Patrick. Sorry for the late reply here. As I previously said, I think the tool sounds useful to me and the pricing seems fair. From looking at the milestone again, I have just one additional request. Could you add docker images, delivery reports (e.g. how to install, compile, run, and/or test the deliverable) as well as unit tests if appropriate to the milestones? Just to make it easier for us to test the deliveries ;-). Also which programming language are you planning to use (python)?

pventuzelo:
Hi David, 
I've updated the milestones ;) 

Regarding the programming language used in the project, I think mainly Rust for fuzzing development and Python for scripting.

PR #12, opened 2020-04-14
propose Java client lib

splix:
# Grant Application Checklist

### DIFF
--- a/applications/java-client.md
+++ b/applications/java-client.md
@@ -38,7 +38,7 @@ Build artifacts are going to be published to a publicly accessible Maven reposit
 
 * **Total Estimated Duration:** 3 months
 * **Full-time equivalent (FTE):**  0.8 
-* **Total Costs:** 4.2 BTC
+* **Total Costs:** 4.5 BTC
 
 ### Milestone 1
 
@@ -46,7 +46,7 @@ Initial domain specific model, types and encoding.
 
 * **Estimated Duration:** 1 month 
 * **FTE:** 0.8
-* **Costs:** 1.4 BTC
+* **Costs:** 1.5 BTC
 
 | Number | Deliverable | Specification | 
 | ------------- | ------------- | ------------- |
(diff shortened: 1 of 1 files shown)


Noc2:
Thanks for the application. It looks really good. Could you maybe add docker images and unit tests?

splix:
@Noc2 sure, I thought unit tests do not need special mention, will add them. Instead of Docker I propose to make a simple Example App, as a code, so anyone can play with it, compile, run.

splix:
@Noc2 added unit tests

gautamdhameja:
Mostly looks good to me too. I would suggest adding a bit more details to Milestone 2 regarding accessing Substrate APIs. Does it cover all RPCs exposed by Substrate node?

splix:
@gautamdhameja that's the tough part. It doesn't seem to be feasible to cover _all_ of the APIs in a single milestone. The goal is to support _most_ of critical APIs. So it's a bit unclear right now, also unclear what is the final list of APIs, I only see them in the implementation right now (would appreciate a link to spec, I guess I missed it). It also makes sense to finalize extrinsic-related APis only with the Milestone 3

gautamdhameja:
Yeah, it makes sense to target sending extrinsics related APIs in milestone 3. In milestone 2, you can target storage reads, system and runtime metadata, and listening to events (subscriptions). 

Here's a list of RPC calls supported by the JS client - https://polkadot.js.org/api/substrate/rpc.html

In the Java client, we might not need all of them right away. For example, at this time, we can safely assume we won't need `engine`, `babe`, and `child storage` categories, I think.

splix:
@gautamdhameja yeah, that's exactly my thoughts. Want to start with the most essential things, and add things one by one when it needed.

splix:
@gautamdhameja @Noc2 I've updated the proposal with RPC details

splix:
Sure, I'll try to do the best I can. Better to overdeliver than overpromise.

PR #16, opened 2020-05-04
RSA Signature Verification on Substrate

yuki-js:
# Grant Application Checklist

### DIFF
--- a/applications/rsa-verification.md
+++ b/applications/rsa-verification.md
@@ -1,8 +1,8 @@
 # Open Grant Proposal
 
 * **Project:** RSA Signature Verification on Substrate
-* **Proposer:** @yuki-js
-* **Payment Address:** TBD
+* **Proposer:** [@yuki-js](https://github.com/yuki-js)
+* **Payment Address:** 12whryLeVFj6ieL6x1jZzC7fFhPfSzfKMh
 
 
 ## Project Description :page_facing_up: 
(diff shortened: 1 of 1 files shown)


burdges:
You'd use which RSA implementation?  https://github.com/briansmith/ring maybe?

You want RSA support in substrate for your project?  What is that project?

We'd need some mechanism for chains exclude, or increase fees for, based upon signature types in `sp_runtime::MultiSignature` because obviously most substrate chains, including polkadot, should disable support for RSA, and chains that support RSA should charge higher transaction fees.  We already need slight fee increases for ECDSA secp256k1 signatures, but not there currently.

yuki-js:
The current implementation of Mynachain uses https://github.com/RustCrypto/RSA . But ring crate  should also be usable.

Mynachain is the blockchain that enables the mynumber card to send the assets. Anyway, I added what the Mynachain is.

>chains that support RSA should charge higher transaction fees

Is it because the signature of RSA is much longer than that of ECDSA and consume much storage space?

burdges:
Yes RSA signatures are much larger.  Also, verifying RSA signature requires far more CPU and they lack batch verification.  

As an aside, Rabin-Williams signatures employ modular arithmetic similar to RSA, but provide much faster verification https://crypto.stackexchange.com/a/564 but they still require large public keys.  There exist other fast signature types with even larger key sizes.  

Polkadot would ban RSA outright of course.  In fact, we'd make banning RSA must be the default configuration, so one interface question is:  How should chains configure `Multisignature`?

I'd expect  https://github.com/RustCrypto/RSA  should be fine.

yuki-js:
RSA verification is much faster than ECDSA verification, isn't it? https://arxiv.org/pdf/1508.00184.pdf

The way to configure `Multisignature` I thought is

1. Macro that generate it
2. Copy the code of `Multisignature` yourself.

Which

burdges:
Yes oops.  It appears equivalent 3072 bit RSA verification runs 3x faster under openssl on my machine:   `openssl speed rsa` and `openssl speed ed25519` report
```
                                              sign              verify          sign/s      verify/s
rsa 2048 bits                        0.000848s  0.000030s  1179.5     33214.3
rsa 3072 bits                         0.002791s  0.000054s  358.3      18480.3
253 bits EdDSA (Ed25519)   0.0001s      0.0002s      16093.0   5949.2
256 bits ecdsa (nistp256)    0.0000s      0.0001s      26078.1   9509.2
```
Except those ed25519 vs ecdsa p256 benchmarks disagree with better code https://www.imperialviolet.org/2010/12/21/eccspeed.html so maybe perhaps that openssl ed25519 implementation sucks.  

It's plausible ed25519 nearly ties RSA 3072 for verification speed with proper implementations for both and batch verification for ed25519.  Anyone using RSA might employ faster 2048 bit keys though. 

Yes, we'd worry about the signature sizes though since equivalent RSA signatures occupy 6 times more space, but yeah maybe sharding will change this.

burdges:
Anyways, there is no reason not to support RSA for parachains because we've oodles more strange zk-crypto that'll require support, so the questions are more like "how do we avoid messing up multisignature for anyone normal?" and "at what layer in the signature do we interact with the host for this?"

yuki-js:
Ok, I'll close this PR once and resubmit when I came up with better interface for signature.

burdges:
We could check if a trait could provide functionality like `Multisignature`.

PR #23, opened 2020-05-15
sr25519-donna

TerenceGe:
# Grant Application Checklist

### DIFF
--- a/applications/sr25519_donna.md
+++ b/applications/sr25519_donna.md
@@ -46,6 +46,7 @@ Please provide the following:
 | 3. | sign/verify message | support signing/verifing messages |
 | 4. | random number generator | provide default random number generator and allow users to use their custom version |
 | 5. | curve operation | take advantage of ed25519-donna's curve operations, support both 32bit and 64bit operations |
+| 6. | add documentation | add documentation of installing or integrating these features |
 
 ### Milestone 2
 
@@ -58,6 +59,7 @@ Please provide the following:
 | 1. | vrf sign if less | support verifiable random function signing operation |
 | 2. | vrf verify | support verifiable random function verifing operation |
 | 3. | unit test | provide unit test for vrf siging and verifing |
+| 4. | add documentation | add documentation of installing or integrating vrf features |
 
 ## Additional Information :heavy_plus_sign:
 Any additional information that you think is relevant to this application that hasn't already been included.
--- a/applications/sr25519_donna.md
+++ b/applications/sr25519_donna.md
@@ -30,13 +30,13 @@ Please provide the following:
 
 ## Development Roadmap :nut_and_bolt:
 
-* **Total Estimated Duration:** 5 weeks
-* **Total Costs:** 1.2 BTC
+* **Total Estimated Duration:** 4 weeks
+* **Total Costs:** 0.6 BTC
 
 ### Milestone 1
 
 * **Estimated Duration:** 2 weeks
-* **Costs:** 0.4 BTC
+* **Costs:** 0.3 BTC
 
 
 | Number | Deliverable | Specification |
@@ -50,7 +50,7 @@ Please provide the following:
 ### Milestone 2
 
 * **Estimated Duration:** 2 weeks
-* **Costs:** 0.4 BTC
+* **Costs:** 0.3 BTC
 
 
 | Number | Deliverable | Specification |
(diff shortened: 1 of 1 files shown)


Noc2:
Thanks for the application. Could you provide a little bit more background/update the project description? Also for anything cryptography related we usually ask teams to provide links to prior work and their general experience in the field.

TerenceGe:
Thanks for the review, just updated some background info and my previous work links in the pr.

TerenceGe:
@Noc2 any feedback so far?

Noc2:
I asked our Jeff to take a look at the application. He will probably comment on it here, once he has some time.

burdges:
What is the background of the person doing the implementation?  How much of the existing new code is home grown?

Why not use libsodium?  https://doc.libsodium.org/advanced/point-arithmetic/ristretto I know one earlier push for a C ristretto implementation based on ed25519-donna got abandoned, but probab ly only due to lack of time, not inconveniences in the code.

burdges:
We've fixed the VRF so it could be implemented, but about the only reason for VRF on iOS seems like games.  We've actually discussed using VRFs for games recently, so maybe..

Is there any immediate need for the mulisig somewhere?  We're still trying to prove one new two round version to be secure, which might simplify the code. Also https://github.com/w3f/PSPs/pull/7 requires I clean up the code considerably before anyone adds some DKG.

TerenceGe:
This library is originated when I build a mobile wallet for polkadot, the core is built in C, so I think it's better to have a C version for the algorithm. I dig into the code of https://github.com/isislovecruft/ristretto-donna, and found its usable after fixing some issues https://github.com/isislovecruft/ristretto-donna/pull/4/files, and I think it's more light weight than libsodium. And for those project using ed25519-donna (such as trezor), it's more convenient to integrate polkadot into their system.

TerenceGe:
mulisig might not require currently, but since the lib's goal is to keep sync with rust version, we can support that once it's ready.

burdges:
I see, so the Ristretto code came from you fixing Isis Lovecruft's implementation?  Very good.  :) 

From where did the merlin code originate? 

Right now, we've this [`add_trusted`](https://github.com/w3f/schnorrkel/blob/master/src/musig.rs#L591) method in musig that violates the extra initial hash commitment's assurances.  We've multiple (pre)witness values confusingly called [`REWINDS`](https://github.com/w3f/schnorrkel/blob/master/src/musig.rs#L59) that [we "delinearize" to produce the real witness](https://github.com/w3f/schnorrkel/blob/master/src/musig.rs#L404).  We actually believe this suffices making `add_trusted` secure making the initial hash commitment unnecessary.  We're making progress on the security proof, but it's really been a slog since nobody could work on this full time, and it'll depend upon non-standard assumptions.  Also, the musig code also needs a cleaner interface to DKGs like I said.  We'd ideally sort this stuff out first before encouraging people to put things in airgaped devices, but the world is not perfect so this all depepends upon who wants what when.  :)

TerenceGe:
The merlin code is based on the author's implementation (https://github.com/hdevalence/libmerlin), also I fixed a little issue and add some stuff.
It's ok that I can remove the muti-sig feature in this application, and just focus on the first 2 mile stores currently. And I can add additional features in the future plan if required.
How do you think about this? If it's ok, I'll update the application.

Noc2:
Just for reference, there is also the following implementation: https://github.com/usetech-llc/sr25519

Noc2:
Could you provide some more information about your iOS app, since I’m currently not sure how useful it would be for our ecosystem. Apart from this, I think it makes sense to remove the third milestone for now. Also, it obviously helps if you could reduce the price a little bit further. Thanks

TerenceGe:
@Noc2 apple's infrastructure dose not support rust very well (https://github.com/rust-lang/rust/issues/35968). Although it's possible to use rust lib as static lib for ios, that requires you disable bitcode, which prevents you from taking advantage of apple's further optimizations (such as reducing bundle size) for you. That's why some famous libraries such as react native dose not use rust currently (https://twitter.com/sebmarkbage/status/1213199159840260096).

Lederstrumpf:
> I dig into the code of https://github.com/isislovecruft/ristretto-donna, and found its usable after fixing some issues https://github.com/isislovecruft/ristretto-donna/pull/4/files, I think it's more light weight than libsodium. For those projects using ed25519-donna (such as trezor), it's more convenient to integrate polkadot into their system.

Looks good to me. Libsodium also has a minimum build option, but ristretto is one of the features excluded there.
We have another grant for a C-library targeting HSMs for key management - the library is currently on libsodium, so it may be of interest to them to switch to a lighter weight ed25519/sr25519 dependency for their embedded devices.

TerenceGe:
@Noc2 Ok, just added documentation into the milestones

TerenceGe:
@Noc2 I prefer the open grants program, BTC is ok for me.

Nittiyadc:
@reto trinkler ?????

Noc2:
@Nittiyadc I deleted the comment, since it’s against our [code of conduct](https://github.com/w3f/Open-Grants-Program/blob/master/CODE_OF_CONDUCT.md) and not related to this grant.

PR #24, opened 2020-05-17
Phragmen wasm

derfredy:
# Grant Application Checklist

### DIFF
--- a/applications/phragmen-wasm.md
+++ b/applications/phragmen-wasm.md
@@ -1,0 +1,78 @@
+# Open Grant Proposal
+
+- **Project:** Phragmen WASM
+- **Proposer:** Colm3na
+- **Payment Address:** 3982fvdPfs1hHUF7BYc2GbaU3LdfN4Ruzg
+
+## Project Description :page_facing_up:
+
+Create a new phragmen implementation for WASM.
+
+This implementation is intended to work as a library, so it can be embedded
+in any application and since it will be a WASM module, it could potentially run
+inside any programming language or runtime.
+
+Unlike [offline-phragmen](https://github.com/paritytech/offline-phragmen), the
+implementation will not require external services
+(like a substrate node). You will be able to pass any input to the algorithm and
+it will just compute the result. This can be useful for block explorers like
+Polkastats which already have all the information needed to run the algorithm
+and currently needs to run `offline-phragmen` binary which fetches again the
+data.
+
+## Team :busts_in_silhouette:
+
+- **Members:** Mario, Fredy, Javi, Diego
+- **LinkedIn Profiles:**
+  - [Mario](https://www.linkedin.com/in/mariopinouceda/)
+  - [Fredy](https://www.linkedin.com/in/alfredoluque/)
+  - [Javi](https://www.linkedin.com/in/javihg)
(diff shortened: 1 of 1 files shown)


Noc2:
Thanks for sharing the draft. It would be nice if you could add some tests to it as well as documentation for each milestone (see the [Milestone Deliverables Guidelines](https://github.com/w3f/General-Grants-Program/blob/master/grants/milestone-deliverables-guidelines.md) of the general grants program). And maybe add one or two sentences to each milestone to explain it a little bit more concrete. Additionally, our finance team might ask you to provide an estimated [FTE](https://en.wikipedia.org/wiki/Full-time_equivalent) for the project.

kianenigma:
I have major concerns about this approach and proposal overall. 

1. Offline phragmen is merely a playground like tool and indeed I never intended to see it is an external tool. For now, it is handy and provides json output in case some tools need urgent access to Phragmen results off chain. So, best not to even compare it here. Even once there exists some implementation of Phragmen in wasm, I intend to maintain it with low priority since it allows me to debug a lot of untested features. 

2. This proposal needs to specify which algorithm(s) it will implement. seq-phragen is what the name is implying, yet I must confirm that we are currently far-away from seq-phragmen. What you see on Kusama, is the output of this pipeline: `seq-phragme -> balacing(equalize) -> reduce -> compact -> uncompact` 

(note that we lose accuracy in the compacting part, hence to get the _exact_ same output, you need to take it into account). 

So if this proposal achieves implementing seq-phragmen, I am worried that you will realize that many tools cannot reliably depend on it, because it will output a different result. 

2.1 We are moving even further away from phragmen. web3 foundation research team recently released a new paper about this which I will link below. 

3. The current crate that contains all the phragmen related code in Substrate is already embedded in the runtime, is `no_std` and hence it is in principle possible to compile it directly from rust to wasm. I have not tried this but I think before making an attempt at writing another version of the same code, effort should be put at this. 

With all that being said, 

1. I indeed still see some merit in having different implementations of phragmen. But if the target is wasm, trying to cross-compile rust version to wasm is a lot more sensible. 
2. Even if the output of the chain is not reproducible, such implementations are still worthy for rough prediction and estimations. 

So all in all, I don't think the proposal is sensible as it is, but I don't want to discard it altogether either and be to negative. It needs better planning and re-thinking about what we want to achieve, and better understanding of the current state of phragmen and everything related.  I would be happy to discuss it further offline. 

--- 

Related links: 

- https://research.web3.foundation/en/latest/polkadot/NPoS/
- The Phragmen replacement by Web3: https://github.com/w3f/research-internal/blob/master/papers/NPoS_validator_selection.pdf 
- Current phragmen pipeline (subject to change based on the above) https://www.crowdcast.io/e/sub0-online/7

Noc2:
The team informed me that for now they won’t continue to work on the application.

retotrinkler:
I'd see value in this - even if it's a moving target with the actual validator selection algorithm as long as its primarily a wrapper it should be versatile enough - and the way I read the situation they would likely maintain it if they use it on their site.

I would ask the team to reconsider - I would approve this grant.

kianenigma:
> as long as its primarily a wrapper around substrate pallet it should be versatile enough

Indeed, this is much more maintainable and reasonable.

PR #25, opened 2020-05-19
Liqum Open Grant Application

denchik37:
# Grant Application Checklist

### DIFF
--- a/applications/Liqum Open Grant Application.md
+++ b/applications/Liqum Open Grant Application.md
@@ -1,14 +1,10 @@
 # Open Grant Proposal
 
-> This document is referenced in the terms and conditions and therefore needs to contain all the required information. Don't remove any of the mandatory parts presented in bold letters or as headlines! See the [Open Grants Program Process](https://github.com/w3f/Open-Grants-Program/blob/master/README_2.md) on how to submit a proposal.
-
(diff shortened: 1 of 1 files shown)


Noc2:
Thanks for the application. If I remember correctly, you previously also applied via the general grants program and were interested in receiving DOTs. If this is still the case, I suggest to focus on the general grants program application, since for us it’s easier to accept grants that are willing to take DOTs. 

Regarding your application, I think it would be very helpful if you focus on a specific use-case and potentially also have some kind of industry partner that would work with you together. I also have concerns regarding legal issues of such an application and how would companies deal with the transparency that this would create.

denchik37:
You are right in saying that we previously applied through general grants program, yet we were not issued one, as reflected in the 'Additional Information' section of open grants application.

While we are early stage, the ambition is indeed to become a parachain due to the cross chain nature of the envisioned functionality, hence the need for DOTs.

I realize that social proof, or better still traction, would help you make a decision. As a bootstrapped project, however, we need to use every existing opportunity to extend our runway.

Can you please elaborate on the transparency concerns, perhaps in a private message? Thanks!

denchik37:
As discussed via email, our team will focus on general grants program instead, so closing this PR

PR #38, opened 2020-07-20
Add proposal for Substrate Storage Inspector

yanganto:
# Grant Application Checklist


Noc2:
Some additional notes from others: It seems most of the functionality described as part of the grant is already provided by polkadot JS. A much more in depth substrate storage navigator, some detection of extra keys that should be cleaned up, unused keys for many blocks, reconsolidation, commonly accessed keys would be interesting for us.

yanganto:
Hi, Noc2

After discussion with our team members,
and check out the features in PolkadoJS.
The proposal is updated.

We want to point out the value of this tool,
and how it can benefit the ecosystem.

What point the tool directly provides
-	visualization of data
-	provide diff on a pallet or on a hash map
-	try to resume the  key with semantic from hash_concat

Besides, here are two using scenarios for example
-	Runtime developers can use this tool to find out problems much quicker than before. (Just like to put breaking point on the blockchain)
-	It can be easier to evaluate a runtime module/wasm effect on the chain

Furthermore, the tool will try to have the least dependency to be a slim tool, such that it can work on a node that is not active, namely without RPC.

Please consider this proposal if our opinions are acceptable.

Noc2:
Thanks for the updates. Could you point out the difference between your solution and polkadot.js inside the application? Also, the application is still quite expensive.

yanganto:
Thanks for your review.  We are redesigning our solution and change our scope and will update here later.

PR #41, opened 2020-07-29
Substrate Graph

2075:
A compact graphql interface for indexed polkadot and substrate based node data.

# Grant Application Checklist

### DIFF
--- a/applications/substrate-graph.md
+++ b/applications/substrate-graph.md
@@ -1,0 +1,73 @@
+# Open Grant Proposal
+
+* **Project:** Substrate Graph
+* **Proposer:** 2075
+* **Payment Address:** bc1qq0pz4r3mehdeqfy9nmextq0pmvku2rukwkgv5z 
+
+## Project Description :page_facing_up: 
+
+* A compact indexer for @paritytech substrate based nodes providing a graphql interface.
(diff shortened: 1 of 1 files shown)


2075:
Hey there, thanks for the reply.

- polkastats is already pretty much feature complete, as it already includes working code for harvesting blocks
- polkadot psql seems to be a mix between indexing/harvesting, without a graphql but a rest interface
- ETL pipeline possibly is similar to polkastats and our proposal in terms of the resulting stack
- open web3 seems very broad (many different features), cannot tell about implementation detail differences
- none seem to feature helm charts or frontends

the first three seem to aim at similar outcomes, while the psql lacks a graphql interface. polkastats and multiblockchain etl seem both to use hasura at the end.

in our case it is generalised as well with a possible focus on example frontends. polkastats crawler/harvester code could be beneficial to achieve that, instead of building the same generalised stack again.

so yes, if you already have 4 pretty similar projects, plus the internal substrate-archiver, if you really need another one.

Noc2:
I will close the PR, because of the amount of similar projects and since it seems nobody else from the committee is interested in the application. Feel free to apply for another grant. I just saw that you also have the swift grant application still open. At least I think we don’t have anyone working on a swift implementation.

PR #64, opened 2020-09-24
Adding proposal for Kusama tip button

edisinovcic:
# Grant Application Checklist


mmagician:
@edisinovcic Thanks for the application.

I have two comments:
- As far as I understand, when proposing a tip, the proposer cannot specify the amount. Only the council members propose amounts once the tip's been created. For a donation (which is just the transfer from proposer to recipient?) this would make more sense. 
- There is a certain amount of deposit required from the tip proposer to submit a tip. It would be worth including this information in the UI.

Please consider the above points. It would be great if you adjust the mockups to reflect it.

@Swader any further thoughts on this?

mmagician:
@edisinovcic Also please sign the Contributor License Agreement.

Swader:
No further thoughts, both those points are touched on in my original RFP and should indeed be addressed in the UI of the button and any popups it summons, yes.

We should also think about mobile strategy. It's not worth including the button at all in regular mobile browsers, but if someone visits the website through a web3-enabled browser on mobile which already contains the PolkadotJS API in some way, the button should be able to reuse that and interact with it. This eliminates added load and lag on mobile sites.

Swader:
I would like to see more justification on the use of ReactJS beyond "familiarity with ReactJS". It has the potential to needlessly bloat the JS size in an ultimately very simple widget.

edisinovcic:
We've updated proposal to reflect above proposals. React will not be used to reduce overhead. On mobile part we can only detect if website has polkadot js api we can show the plugin, if not, plugin will not be shown.

edisinovcic:
On the cla part it says to me that I already signed it, but it's not showing it here.

cla

Swader:
Skeptical on Svelte but otherwise seems good!

mmagician (changes requested):
Regarding design & user testing:
In general we consider this as the team's own input (R&D) and not fund it through the milestones.
It seems like you already have some mockups.
How were you planning to conduct user testing, btw?

edisinovcic:
@mmagician idea for the user testing was to get on a call with a couple of potential users and see their reaction and feedback if this is the UX/UI is good, otherwise, we would update it depending on their feedback.

edisinovcic:
> Skeptical on Svelte but otherwise seems good!

Svelte has a much lower footprint than React. Building this in plain javascript means that we would need to "reinvent" state management, which would potentially lead to much bigger problems than using a solution like svelte. With all pros and cons, we concluded that svelte is least heavy solution.

edisinovcic:
> Skeptical on Svelte but otherwise seems good!

Svelte has a much lower footprint than React. Building this in plain javascript means that we would need to "reinvent" state management, which would potentially lead to much bigger problems than using a solution like svelte. With all pros and cons, we concluded that svelte is least heavy solution.

Swader:
>  "reinvent" state management

Which state management? A `k=>v` store in localstorage is enough to remember whether or not a user interacted with the button, no? What other state management is there in a button which purely sends out an extrinsic and remembers if it did it? Or did you think of adding more information into the storage to make the button's display richer?

dark64:
If you check the mockups we provided, you can see there is a multi step flow going on. Any kind of state management is a must for cases like this, for keeping track of what screen is currently active, what account is currently selected etc. It's not just "is button clicked or not". I understand your concerns, but from my experience with Svelte, the output is minimal and compiled down to vanilla javascript, thus potentially making a smaller footprint than writing this in vanilla js.

Swader:
Okay, understood 👌

mmagician:
> In general we consider this as the team's own input (R&D) and not fund it through the milestones

@edisinovcic What I meant is that we will not fund this. It seems that the original core development was planned for 0.5 BTC (excl. design + testing), while now you've split it up over two milestones & it sums up to 1 BTC. I understand that you might still need to polish some UI & perform user testing, but can we keep the price a bit lower (0.75 BTC)?

Otherwise looks good & we could proceed.

edisinovcic:
@mmagician ah yeah, we understood you incorrectly. Definitely will lower a price to 0.75 BTC.
Tnx for the quick response.

mmagician (approved):

Noc2 (changes requested):
Thanks for the proposal, but did everyone who contributed to the proposal sign the Grants Terms and Conditions?  I think @anamst still needs to sign it.

Noc2 (changes requested):
Just two more additional requests from my side: It currently says total costs: 1 BTC, but milestone 1 plus 2 are only 0.75 BTC. Also could you integrate some article or tutorial as part of the deliveries.

anamst:
Changed the price and added article/tutorial on the deliveries. Is it ok now?

Noc2:
Looks good, but why is it a separate milestone without any costs associated with it? Could you integrate this into milestone 2?

Noc2 (approved):

mmagician (approved):

Swader (approved):

laboon (approved):

anamst:
The work on this grant has been concluded. It has been extensively tested and already deployed at quakeaid.ubik.hr for users to donate to quake relief fund in Croatia. You can see the walkthrough videos below, including the screenshot of the tipping screen (we have modified the text after some initial user testing to make the tipping process clearer). Source code is public and can be accessed on the link below. Please inspect the work and give your feedback. Thank you.

https://github.com/Shard-Labs/kusama-tips-widget
https://user-images.githubusercontent.com/48686305/103631181-f3b39300-4f42-11eb-8da7-0ae3c0bb8422.mov
https://user-images.githubusercontent.com/48686305/103632173-345fdc00-4f44-11eb-99ee-4670052b0494.mov
Tipping screen

Swader:
I would add a warning here that if the Council decides to cancel the tip (a new feature), the bond will be slashed. Maybe finish with something like "Only open tips for content your truly believe add value to the ecosystem"

anamst:
Tipping screen with changed warning is attached. Please have a look and let us know if it's better now.
Screenshot 2021-01-05 at 13 03 18

mmagician:
@anamst Thanks for the update.

I tried it quickly on https://quakeaid.ubik.hr/, but I am unable to click on the `Propose Tip` button, it feels like it's a text-only element and not a button. 

Just a reminder to submit your delivery by following the instructions [here](https://github.com/w3f/Grant-Milestone-Delivery) once you're finished & satisfied with your work.

Swader:
Tipping is disabled on quakeaid, only direct donations work. We felt it would be weird to be able to tip for that.

anamst:
@mmagician You are correct, Propose tip doesn't work on https://quakeaid.ubik.hr/ because we disabled it. It is enabled by default but users have the option to disable it if they think there's a reason to do it. For example, we did it here to keep it simple and uniform with donating DOT

mmagician:
Yeah totally makes sense. Just a tip (no pun intended): maybe hide it from the UI, if it's disabled? It might otherwise be slightly misleading to the end-users, since they can't do anything with it anyway.

dark64:
> Yeah totally makes sense. Just a tip (no pun intended): maybe hide it from the UI, if it's disabled? It might otherwise be slightly misleading to the end-users, since they can't do anything with it anyway.

Thanks for the suggestion! We made some changes to the UI, it should be a better UX now for the end-users.

PR #75, opened 2020-10-22
New proposal: Polkashots.io, a website hosting Polkadot snapshots

nicolasochem:
# Grant Application Checklist

### DIFF
--- a/applications/polkashots.md
+++ b/applications/polkashots.md
@@ -1,0 +1,103 @@
+# Open Grant Proposal
+
+* **Project:** polkashots.io
+* **Proposer:** Nicolas Ochem
+* **Payment Address:** bc1qu45ws52gkjgmwgu8gmqh2ejkcf4clt93jnnvsm
+
+## Project Overview :page_facing_up: 
+
+### Overview
+
+Implement, release and maintain a snapshot website for Polkadot and Kusama.
+
+### Project Details
+
+Substrate has an `--export-chain` and `--import-chain` feature to allow node operators to sync faster.
+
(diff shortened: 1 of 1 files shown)


BenWhiteJam:
Been looking for something like that, a few questions:
- Where will those get stored?
- How often will they get updated?
- Polkadot & Kusama?
- How do you create the signature to verify the validity of the snapshot?

nicolasochem:
* snapshots will be stored on a public google cloud storage bucket
* the snapshot generation frequency will be configurable. 24 hours frequency seems reasonable. Nodes typically sync very fast from a 24 hour lag.
* We will be supporting Polkadot and Kusama.
* we can generate sha256sum of the files, so users will have a proof of valid download. Would you like that? I can add this to the requirements if needed.

However, the best way to ensure safety IMO would be to provide an option to `substrate import-blocks --hash 0x37d2713b2a53...` to be able to pass the block hash that you think you are importing. You can match that with what you see on the snapshot website.

Noc2 (changes requested):
Sorry for the late response. This is indeed an interesting application. Could you make the tutorial part of the deliveries. Also could you integrate your reply above in the requirements? And in general define the requirements/milestone a little bit more specific (exact functionality, programming language, framework, etc.)?

nicolasochem:
@Noc2 thanks! All done.

Noc2 (approved):
Looks good to me. Thanks!

mmagician (approved):

laboon (approved):

PR #93, opened 2020-11-11
Application: Dinfra

vonhenry:
# Grant Application Checklist

### DIFF
--- a/applications/Dinfra.md
+++ b/applications/Dinfra.md
@@ -1,0 +1,256 @@
+# Open Grant Proposal
+
+* **Project:** Dinfra
+* **Proposer:** vonhenry
+* **Payment Address:** 327ZupUbnFbHAsZVPdKBLmtFPnKdr1Knyv 
+
+## Project Overview :page_facing_up: 
+### Overview
+* A brief description of the project.  
+  The goal of this project is to provide instant, scalable API access to the Polkdot, Filecoin and IPFS networks, including blockchains' historical data.
+
+  For DApps, stable, reliable and rich node API services are very important. However, it is not a simple thing to run multiple kinds of  blockchain nodes, blockchain historical data retrieval services, and ensure the high availability of each service, especially when DApps use a variety of underlying technologies and blockchains (such as Filecoin and IPFS), the operation and maintenance of infrastructure will be heavy work. The purpose of this project is to provide fast and highly available API services.
+
+* An indication of how you will integrate this project into Substrate / Polkadot / Kusama.  
+  Substrate and Polkadot provide a powerful parachains framework and ecosystem, Filecoin and IPFS provide a perfect distributed storage scheme. The combination of the two will support more diverse and complex decentralized applications.
+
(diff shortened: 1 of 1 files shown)


Noc2 (changes requested):
Thanks for the application. However we require more details regarding the exact implementation (e.g. programming language and functionality) For example regarding account and management system, will it support polkdato as well filecoin accounts, in which programming language will it be implemented, how is tested (unit-tests), etc. Also teams should usually have some initial specification of the project before they apply for a grant. Additionally p2p is currently working on a Multiblockhcain ELT (filecoin, polkadot): https://github.com/p2p-org/mbelt-docs How is your solution different from their implementation?

vonhenry:
Thanks for your reply, for your questions and my answers are as follows:

1. regarding account and management system, will it support polkadot as well filecoin accounts?  
Account system refers to the user's account registered on the website, which is used to manage the user's service, balance and other information. The account can contain multiple blockchain accounts of the user, including Polkadot and Filecoin accounts, as required.

2. in which programming language will it be implemented, how is tested?  
We use Golang to implement the main parts of Dinfra. 
In each repo, we will add enough unit tests, and we also have cluster tests.

3. Also teams should usually have some initial specification of the project before they apply for a grant.   
Please refer to https://github.com/Dinfra/specs

4. https://github.com/p2p-org/mbelt-docs How is your solution different from their implementation?  
There are many similarities between these two projects, but we focus on Polkadot and Filecoin ecosystem, We will provide users with as many tools and convenience as possible. For example: 
- Filecoin excellent miners details and credit list
- One click selection of the most suitable filecoin miner for a user
- Data encryption storage
- Convenient exchange token of DOT and FIL
We believe that the combination of the two is enough to support almost any type of DApps, and we provide solutions that can be implemented commercially, we can obtain revenue by charging users service fees, and form a positive business logic. We have rich experience in blockchain research and development, and we believe that we will do the best.

Noc2:
Thanks for the update. Could you integrate this information as well as the existing specification into the grant application and update it accordingly? The more technical details you can provide the better.

vonhenry:
Above information and the specification have been merged into `Dinfra.md`,  and development roadmap has also been updated.

mmagician:
@vonhenry Thanks for the application, I have a couple of comments:

1. Could you please double check the consistency of naming in your application? I'm not sure whether `promisor` & `promoter` are supposed to be the same or different components.
2. What data exactly are you going to store in the reversible & irreversible databases? Will there be continuous data movement between these two, and if so, at which point do you decide that data is safe to move to the irreversible DB?
Also, perhaps you could talk about how Polkadot consensus mechanism affects the finality of your data.
3. Could you list all the libraries that you're planning to integrate into your solution? (perhaps the [Go Substrate RPC Client](https://github.com/centrifuge/go-substrate-rpc-client))
4. > Convenient exchange token of DOT and FIL 

This will be a centralised service, correct? Can you elaborate please?

5. From what I can see, when you compare your solution to [Multiblockchain ETL](https://github.com/p2p-org/mbelt-docs), most of the points from the list regard Filecoin & not Substrate (and please keep in mind this is a Polkadot ecosystem focused grant). I like the idea that a user could pay in DOTs for using your service, perhaps you could add more details on how this solution would be implemented in practice?

PR #111, opened 2020-11-19
Automate the application acceptance process

mmagician:
the only drawback is that the startRow needs to be updated periodically (ATM every ~25 accepted applications)


Noc2:
Let’s see if it works. I hope you made a copy of the sheet ;-)

PR #112, opened 2020-11-20
Stingray-SNARK Proposal

stechu:
# Grant Application Checklist


mmagician (changes requested):
@stechu Thank you for the application, we are indeed interested in developing a ZK framework for Polkadot.
I have a few comments regarding your application:

1. What exactly are the new functionalities that you would like to implement? There are a number of existing Groth16 Rust implementations already (e.g. https://github.com/arkworks-rs/groth16). Are you planning to fork one of these libraries and build on top of it? If so, could you please mention that explicitly in your application, together with the concrete list of extra features that you'll add there?
2. We don't generally support testing as a separate milestone, but rather require it to be delivered along with each milestone. Therefore, could you please include it as part of M1?
3. What are you going to benchmark against?
4. For Milestone 2.2, could you elaborate on what you mean by the `3 different use case examples` please?

stechu:
@mmagician, thanks for the review! Please see our revised grant application more specific and addressed your comments:

1.  Part of our implementation will be based on the arkworks code. This specific grant application would provide two Rust crates that is similar to `arkworks::algebra` and  `arkworks::curves`. The major work that we are going to do are two parts:
     * Refactoring `arkworks::algebra` and `arkworks::curves` code (including reimplement the part that requires std)  to make it runnable on substrate.
     * Schnorr Signature, BLS Signature, Pedersen based on `stingray::algebra` and `stingray::curves` 
 we have updated the description in the proposal
2. Please see the revised milestone.
3. We would imagine stingray crates will be slower than arkworks counterparts due to lacking support of AVX (as far as I know) when compile rust to WASM. So we will benchmark against the arkworks counterparts.
4. We provide a more specific milestone in the revised proposal.

burdges:
Any grants by W3F for the elliptic curve topics listed here should pay for _upstream merges_ into crates controlled by Arkworks, W3F, Parity, or MaterLabs, or perhaps established projects like halo2.  Arkworks actively engages and accepts pull requests, so supporting perma-forks harms the zk ecosystem. 

We've an internal BLS signature project for bridges, not currently public.  We'd appreciate it if grants avoided BLS until released, and after that restricted itself to improvements to our own crates.

stechu:
@burdges  thanks for the feedback! We didn't know there is W3F effort on implementing BLS. The existing W3F [BLS project](https://github.com/w3f/bls/blob/master/src/verifiers.rs) doesn't seem can be used in substrate since it also required std.

We are very happy to engage with upstreams such as ArkWorks. The reason that we were proposing a fork is that the code change is non-trivial, if we want to port arkworks::algebra and arkwork::curves to **substrate constrained Rust**.  And we are uncertain ArkWorks attitude of adding substrate support. Actually, one of our design goals is to be compatible with upstream so that in the future the library could benefits from upstream improvement. So it is good to know they are very open about that.

As for this grant proposal, do you have any concrete suggestion? If there is near plan that arkworks team will add substrate constrained Rust support, we are happy to retreat the proposal and just use that in our project. If not, we are also happy to change the grant proposal to the forms of PRs to arkworks upstream repos. 

We will remove the BLS part of the proposal and wait for W3F's official BLS.

burdges:
Arkworks expects considerable churn for the coming year or so, so it's hard to stay compatible unless you upstream changes.

> We want to port arkworks::algebra and arkwork::curves to substrate constrained Rust

Arkworks already works without std, and provides non-asm fall backs, so what issues remain?  

If I recall, there is already a zcash sapling fork that used zexe so that it could be ported to substrate, but maybe that project stalled.

stechu:
make sense. Indeed, I double checked ArkWorks code. Now, only x86-64 backend requires `std`.
thanks for the review.

PR #114, opened 2020-11-21
Application: Konomi 

debbiecho:
# Grant Application Checklist


mmagician (changes requested):
@debbiecho Thank you for an interesting application!
1. I didn't quite understand how the system guarantees that a borrower has enough collateral. Specifically, using the `borrow` callable from the lending module, there is no way to supply parameters that allow specifying which assets are used as collateral. Therefore, when I call `borrow`, how does the module know if I am allowed to actually borrow?
2. Can anyone call liquidate? If the conditions for liquidation are met, does the caller (the arbitrager?) receive the liquidated amount, or is it burnt? Where does the bonus come from?
3. When I look at the `amount`, `index` & `as_collateral` fields that will be stored for each user, it seems like this only concerns one asset. Therefore, do you need to store these fields for each asset that a user interacts with?

Thanks upfront for your responses!

debbiecho:
> @debbiecho Thank you for an interesting application!
> First of all, please sign the CLA, this is necessary condition for reviewing your application. Also, I have a few questions:
> 
> 1. I didn't quite understand how the system guarantees that a borrower has enough collateral. Specifically, using the `borrow` callable from the lending module, there is no way to supply parameters that allow specifying which assets are used as collateral. Therefore, when I call `borrow`, how does the module know if I am allowed to actually borrow?

Actually all the borrower's assets he/she enables as collateral are used as collaterals. For example, user deposit $100 BTC, $50 ETH and $200 DOT, and enables BTC and DOT as collaterals (if the system allows), then $100 BTC and $200 DOT together count for user's total collateral. Therefore, when user's total collateral amount is below liquidation threshold, arbitrager can pay the target token and choose to get either BTC or DOT.


> 2. Can anyone call liquidate? If the conditions for liquidation are met, does the caller (the arbitrager?) receive the liquidated amount, or is it burnt? Where does the bonus come from?

Yes. And the caller gets the liquidated amount as an incentive to keep the system healthy. About the bonus, because when user borrows, he/she needs to over-collateralize (say he/she needs 150% equivalant value of the assets he/she wants to borrow). And liquidation threshold is also above 100%, say 120%. This means when liquidation happens, user's collateral worthes 120% of the assets he/she has borrowed. In this case, the bonus comes from the 20% additional value.

> 3. When I look at the `amount`, `index` & `as_collateral` fields that will be stored for each user, it seems like this only concerns one asset. Therefore, do you need to store these fields for each asset that a user interacts with?

Yes, for each asset or pool, we need to track his/her amount and index for debt and amount, index & as_collateral for supply.

> Thanks upfront for your responses!

debbiecho:
@mmagician

mmagician:
@debbiecho Thanks for your answers, they were on the point. I would be happy to go ahead with this application.
Just one more question regarding the front end - are you planning to build your own website, or fork & adapt [the frontend template](https://github.com/substrate-developer-hub/substrate-front-end-template)?

debbiecho:
> @debbiecho Thanks for your answers, they were on the point. I would be happy to go ahead with this application.
> Just one more question regarding the front end - are you planning to build your own website, or fork & adapt [the frontend template](https://github.com/substrate-developer-hub/substrate-front-end-template)?

Thanks @mmagician for the quick feedback. We are starting with the frontend template now but we will switch to our own website for more functionalities.

mmagician:
@debbiecho sounds good. In this case could you please clarify what is the scope of the milestone, i.e. is your own website already part of the grant or not.

debbiecho:
@mmagician yup, i just updated the application doc! To clarify, we already have our own webpage for the swap function, and we plan to build our own website for the lending module as well and that is part of the grant :D

mmagician (dismissed):

Noc2 (changes requested):
Thanks for the application. I have a few follow-up questions: Could you also define the features as well as the programming language/framework of the frontend a little bit better as part of the deliverables and could you potentially integrate a tutorial? It would also be nice if you could take a look at the existing [tech stack](https://github.com/w3f/General-Grants-Program/blob/master/grants/polkadot_stack.md) and point out difference to other protocols. Potentially you can also reuse some of their work.

debbiecho:
> Thanks for the application. I have a few follow-up questions: Could you also define the features as well as the programming language/framework of the frontend a little bit better as part of the deliverables and could you potentially integrate a tutorial? It would also be nice if you could take a look at the existing [tech stack](https://github.com/w3f/General-Grants-Program/blob/master/grants/polkadot_stack.md) and point out difference to other protocols. Potentially you can also reuse some of their work.

@Noc2 yup, we use React framework with Javascript for the front end. The main functions are to allow users to view their account, deposit into different asset pools and borrow with collateral. We are happy to integrate a tutorial and would appreciate more instructions on this if possible (like what to include exactly) In terms of the existing tech stack, we believe that we are offering something new because there is no project covering pool based lending yet but we do benefit from other projects in the broader DeFi category.

Noc2:
Could you also integrate the functionality of the front-end into the deliverables?  Regarding the tutorial, it would be nice if it includes the front-end, either about the development of the front or the interaction with it. I think [acala](https://github.com/AcalaNetwork/Acala) has some lending related implementation for their stable coin. Maybe you could just quickly point out the differences.

debbiecho:
@Noc2 Front end functions integrated into the milestones. Yup, we would be happy to prepare the tutorial about our front-end development. With regards to the differences from Acala, they are mostly using the collateralized debt positions to mint stablecoins, while for our product, we are building pool lending protocol to support multiple assets.

Noc2 (approved):
Looks good to me.

mmagician (approved):

RouvenP (approved):

PR #135, opened 2020-11-30
Automate replies when details are missing

mmagician:
github-action will automatically comment when a PR is labelled as 'details missing'


Noc2:
Looks good

PR #136, opened 2020-11-30
Application: Subquery 

ianhe8x:
# Grant Application Checklist


jamesbayly (dismissed):

mmagician (changes requested):
@ianhe8x Thanks for the application. I have a couple of questions:
1. What would a Subquery project comprise?
2. What do you mean by `compile and pack` the project?
3. You list an indexer as a prerequisite, but then you go on to say in Milestone 1.2 that your node will index the data. So could you please clarify whether the indexer is part of your grant or not?
4. `will execute mapping functions in a sandbox executor one way or another.` ???
5. How do you plan to `make our best effort to keep the indexing result deterministic.`? We will need a little bit more detail than this. Please conduct appropriate research on how/when Polkadot blocks can be considered finalised and at which point they should be indexed.
6. What is the difference between deliverables of milestones 1.1 and 2.1?
Thanks!

ianhe8x:
> 1. What would a Subquery project comprise?

A subquery project is a nodejs project including a project manifest file, graphql schema and user's custom mapping functions.
> 2. What do you mean by compile and pack the project?

Subquery project is written in typescript so we will need to compile it into js. And then perform `npm pack` to put all parts into a tarball file. In this proposal, the `@subql/node` will be given the path to this file and load it from local disk. And in the future we will support other ways of transmission.

> 3.  You list an indexer as a prerequisite, but then you go on to say in Milestone 1.2 that your node will index the data. So could you please clarify whether the indexer is part of your grant or not?

Indexer is not a prerequisite but an archive node is. I will update the proposal to make it more clear.

> 4. will execute mapping functions in a sandbox executor one way or another.

We are looking into several sandbox approaches. WASM is a mature way, though it is very challenging to make `@polkadot/api` available in AssemblyScript.  The promising way at the moment is to use vm2 and run `@subql/node` in k8s with Network Policy configured. I will state this approach explicitly in the proposal to avoid  ambiguity.

> 5. How do you plan to make our best effort to keep the indexing result deterministic.? We will need a little bit more detail than this. Please conduct appropriate research on how/when Polkadot blocks can be considered finalised and at which point they should be indexed.

We are aware of the finality mechanism, `@subql/node` will only feed finalised blocks to the mapping functions. The statement you pointed out is refer in particular to storage queries within mapping function. `@polkadot/api` access is one essential feature we are keen to keep in the proposal, but to avoid user querying latest states instead of state of the specific block under process, we will make some patch to the instance of api before injecting it into executor's context for mapping functions to use, which will lock the storage query to a block without user to use `.at(blockHash)`. (which makes it effective for derive queries as well)

> 6. What is the difference between deliverables of milestones 1.1 and 2.1?

In milestone 2.1 we plan to make improvements on the ease of use. Comparing with 1.1, creating a scaffold is not that essential to the whole user flow.

mmagician:
@ianhe8x Thank you for a clear response. I have a few additional follow-ups:
1) It would also help if you could be a bit more specific on explaining the scaffolding in M2.1 - it should be detailed enough that we can later assess whether the functionality of your delivered code meets the milestones.
2) In addition, perhaps you could elaborate on the handlers from M1.2.1 and how one could make use of them?
3) We've got [another project](https://github.com/p2p-org/polkadot-profit-transformer/tree/master/streamer) developing an ETL solution for Polkadot (it's currently under development). While the documentation is still somewhat sparse there, perhaps you could have a look into what they were proposing and try to identify how you plan to stand out? I can also put you in touch with their team if you'd like, see whether you could collaborate?
4) Assuming the user leverages an existing archive node, what are the computing resources required for running the `@subql/node` by the user? I assume that since an archive node is required, the node will index historical data, right?
5) >Subquery project is written in typescript so we will need to compile it into js. And then perform npm pack to put all parts into a tarball file. In this proposal, the @subql/node will be given the path to this file and load it from local disk. 

Would you mind including the above in the deliverables also, please?

6) Please also include unit & integration tests in the deliverables.

ianhe8x:
More details are added for M1.1 and M2.1 

We are not meant to build another ETL solution, explanation added in the proposal.

Requirements for running `@subql/node` are updated. Archive node is not always required depends on actual use case, even though it is not hard to get access to, people can use the free service we onfinality provide.

Add testing as deliverables, I thought they are implied but never mind now.

mmagician:
@ianhe8x Thanks for the updates.
What I meant with:
> In addition, perhaps you could elaborate on the handlers from M1.2.1 and how one could make use of them?

was to explain how this works. So just by seeing a name `event handler`, I would imagine that a user defines some Substrate-recorded event (e.g, transfer) and defines some callback function to be triggered when such an event occurs. So am I right in understanding it? If so, how can such callback functions be supplied? Are they pre-defined or user-defined?

> Add testing as deliverables, I thought they are implied but never mind now.

Thanks. The application looks good to me.

mmagician (dismissed):

Noc2 (changes requested):
Looks good. Could you reduce the price a little bit more. It’s currently above the 30 k limit? Also could you integrate tests for the second milestone?

ianhe8x:
@mmagician Thanks for the feedbacks. Yes, those handlers are user defined and be packed from the SubQuery project that user created. I've update the statement to make it more clear.

@Noc2 Thanks for the feedbacks. I've update milestone 2 and the total cost.

Noc2 (approved):
Thanks for the update. Looks good to me.

mmagician (approved):
Thanks for the updates, I'm happy to go ahead with it

BenWhiteJam (approved):
Looks good to me as well. Looking forward to it!

PR #153, opened 2020-12-14
bitCheck

bitcheck:
# Grant Application Checklist

### DIFF
--- a/applications/bitCheck.md
+++ b/applications/bitCheck.md
@@ -1,0 +1,110 @@
+# Open Grant Proposal
+
+> This document is referenced in the terms and conditions and therefore needs to contain all the required information. Don't remove any of the mandatory parts presented in bold letters or as headlines! See the [Open Grants Program Process](https://github.com/w3f/Open-Grants-Program/blob/master/README_2.md) on how to submit a proposal.
+
+* **Project Name:** Name of your project (e.g. DuoSwap Module)
+* **Team Name:** Legal name of your team (e.g. Duo)
+* **Payment Address:** BTC payment address. We don't accept payments for the program in other currencies at this stage. (e.g. 123mp123...)
+
+*The above combination of your GitHub account submitting the application and payment address will be your unique identifier during the program. Please keep them safe.*
+
+## Project Overview :page_facing_up: 
+If this application in response to an RFP then please indicate this on the first line of this section.
+
+### Overview
+
+Please provide the following:
+  * A brief description of the project.
+  * An indication of how you will integrate this project into Substrate / Polkadot / Kusama.
+  * An indication of why your team is interested in creating this project.
+
+### Project Details 
+We expect the teams to already have a solid idea about the project's expected final state.
+
+Therefore, we ask the teams to submit (where relevant):
+* Mockups/designs of any UI components
+* API specifications of the core functionality
+* An overview of the technology stack to be used
+* Documentation of core components, protocols, architecture etc. to be deployed
+* PoC/MVP or other relevant prior work or research on the topic
+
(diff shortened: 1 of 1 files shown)


Noc2 (changes requested):
Thanks for the application. This looks quite interesting to me. The only concern I currently have is that milestone 1 is currently quite expensive and doesn’t provide too much value for our ecosystem, given that there are already similar solutions as substrate modules and it shouldn’t be too much work to adapt these. Could you reduce the price of milestone 1 significantly?

bitcheck:
Hi Noc2, we've re-checked the proposal and found these has a calculation mistake. 
Actually we are going to arrange 2 FTEs for Milestone 1, it will cost 0.5BTC(Not 0.75 BTC). And 3FTEs for Milestone 2, it will cost 0.75BTC(not 0.5BTC). 
After we discussed your suggestion and reduced price for milestone 1, That is the new budget:
MS 1: 2FTEs, 0.4BTC
MS 2: 3FTEs, 0.75BTC
MS 3: 2FTEs, 0.25BTC
Ttl: 1.4BTCs

Noc2:
That’s great, but according to the current market price of BTC this would still be above the 30k limit. Could you reduce it a little bit more. Also could you update the price directly inside the application, so I can immediately share it with the rest of the team?

bitcheck:
Okay, application has been updated according to the current BTC price.

Noc2 (changes requested):
Thanks for the update. Sorry I have a few more follow-up questions: 

- Why do you link to the following twitter account on your homepage: https://twitter.com/wix 
- On what is your zk privacy solution based on? Which project did you potentially reuse for it and is it audited? 
- And what’s the purpose of the DividendPool?

bitcheck:
Hi,
1- Our webpage is built by wix.com for the time being. All our team is focus on the app development now, later we will move to standalone web server. Maybe this link is for promotion purpose by wix :)

2- The zk solution is base on Tornado project. We want to improve it on bitCheck polkadot version.
a- Using PLONK instead of GROTH16 to avoid trusted setup while updating circuits. This will improve the expansibility of DAPP.
b- While we use tornado, front-side calculation of generation proof is too heavy to run on mobile wallet. This will be improved also.
c- Tornado is using classic merkle tree to store/track data onchain, but the cost is so high. We will use new protocol like Shrubs to improve it.
Some functions above is realized on bitCheck Ethereum version, but it will be greater to realize on parachain.

3- For attracting users on Ethereum chain and making up for high gas cost, we made this economic model. If the users stake tokens in the DividendPool contract, they will share the payment fee. But we don't think it is necessary on parachain, if the transaction cost become acceptable.

Noc2:
Thanks for the quick reply. Could you integrate this information (except wix.com) into the application? And in general feel free to add as much technical details for each deliverable as possible. This makes it always easier for us to accept an application.

bitcheck:
Hi, the application has been updated with those informations.

Noc2:
Thanks for adding the additional information, but I actually meant could you update the specification of the deliverables in the table to be more precise, for example “zk-knowledge - Submodules: Verifier/ProofGenerator etc.” What is etc.?  Sorry for the confusion. Also what’s your experience with zkps so far?

bitcheck:
Hi, David, Merry Christmas, we have updated the application which have just added teams members' backgroud to explain our experience. But we still need time to discuss the deliverable details to be sure that everything will be performed on time. Maybe tomorrow can be updated for this matter.

mmagician:
@bitcheck I'm closing this for now, it seems that your application is still missing some details. Feel free to re-open the PR once your proposal is complete.

PR #209, opened 2021-01-19
API Protocol

AlexDevAPI:
# Grant Application Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@
 
 ### Introduction
 
-The Open Grants Program seeks to allocate resources to projects that are beneficial to the advancement of the Polkadot ecosystem. It tries to achieve this in the following way:
+The Open Grants Program focuses on funding software development and research efforts related to **Polkadot, Kusama and Substrate**. It tries to achieve this in the following way:
 - Fully transparent
 - Simple
 - Effective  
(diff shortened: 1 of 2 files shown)


Noc2 (changes requested):
Thanks for the application. However, the application is currently above the 30k limit of initial grant applications. Furthermore do you have any technical specifications or code, on which you already worked? You probably want to implement something similar to https://api3.org/, correct?

AlexDevAPI:
> Thanks for the application. However, the application is currently above the 30k limit of initial grant applications. Furthermore do you have any technical specifications or code, on which you already worked? You probably want to implement something similar to https://api3.org/, correct?

I don't think we are the same as API3.org, API3.org is more about helping blockchain contracts to use off-chain data, somehow listing services for contracts, our starting point is to help developers to develop Dapp faster, if I say similar, it may be a bit similar to keep3r network, similar to Web 2.0 enterprise services rapidapi similar, we can apply for less than 30K, currently our project is still in the perfect and team components.

Noc2:
Thanks for updating the price. However, we would still need to have more technical details (programming languages, specification, potentially some initial code, etc.).

AlexDevAPI:
> Thanks for updating the price. However, we would still need to have more technical details (programming languages, specification, potentially some initial code, etc.).

OK, I will organize it, and then submit the PR

PR #211, opened 2021-01-20
Minterest lending protocol application.

DenisRomanovsky:
# Grant Application Checklist


Noc2 (changes requested):
Thanks for the application. However, the application is currently above the 30k limit of the open grants for initial applications. We usually start with smaller grants and issue follow-up grants. Also it would be nice if you could compare your solution to [existing similar solutions](https://github.com/w3f/General-Grants-Program/blob/master/grants/polkadot_stack.md). It probably makes sense to initially only do this for the parts, which you want to integrate into the smaller grant application.

DenisRomanovsky:
@Noc2 , while I am working on comparison with similar solutions, could you please help me with the costs? 

We described our Roadmap in whole to show the final state of the project. 4 out of 6 milestones are subject to either another grant applications or  fund raising. And total costs is calculated roughly based on 6 milestones. The initial grant application was intended to cover only the first two milestones.

How exactly should I fix this? Remove the 4 milestones or set their costs to 0 and tweak the total costs?

Noc2:
Hi. I suggest removing the milestones from the Development Roadmap and instead place them under “Future Plans” to make it clear that they are not part of this grant. Also you should update the total costs accordingly and probably remove the price of the future milestones from application. The bitcoin price might change quite a lot anyway.

DenisRomanovsky:
@Noc2 , thank you for the advice. I've updated the PR with a better Ecosystem fit passage and rearranged the milestones.

Noc2 (changes requested):
Thanks for the update. Could you also update the total estimated duration? Do you already have some kind of whitepaper? Your website currently doesn’t load for me. Also something you should be aware of, we probably won’t support another ethereum bridge at this stage, given that we already have quite a few teams working on it. Including https://github.com/Snowfork/polkadot-ethereum. Also what do you mean by DOT bridge?  Maybe the following project might also be interesting for you: https://github.com/y2labs-0sh/definex

DenisRomanovsky:
@Noc2, updated total duration to the time frame of two milestones, not the whole project. 
We can provide a whitepaper. You need a link or should I send it as a file or something else? 

I've also changed a bit the wording in Plans section. We are not going to build any bridges. The plan is to integrate the solutions, built by other teams, like Plasm or the mentioned polkadot-ethereum.

Noc2:
Makes sense. It would be nice if you could integrate a link to the whitepaper into the application. This could potentially also be used as a reference for your implementation.

DenisRomanovsky:
Understood. Whitepaper is in progress \, the final version should be in place in a couple of days.

DenisRomanovsky:
@Noc2 , I've added a link to a litepaper. Took us more time then expected. A whitepaper is in progress. And we are moving fast towards closing the first milestone.

Noc2 (changes requested):
Thanks for sharing the lite paper. In general something you should be aware of is that the price is currently above the 30k limit. Also we recently changed the template and ask teams now to denominate the project in USD. So usually the payment in DAI is easier than BTC.  Could you still provide more details regarding your implementation and the deliveries? For example how exactly will the pool module work? What is the purpose of the math module? Is it similar to the following: https://github.com/equilibrium-eosdt/substrate-fixed In general how is your solution different from the one developed by https://equilibrium.io/en? How are you calculating the interest rate?  Do you have a mock-up for the UI?

semuelle:
Hey @DenisRomanovsky, is the update in relation to [David's comment](https://github.com/w3f/Open-Grants-Program/pull/211#pullrequestreview-591301677) above? Could you also answer the questions posed?

DenisRomanovsky:
@Noc2 @semuelle , yes, the changes are according to David's comment , plus some tweaks and updates done to make this look closer to what we have in code repos. The only uncovered topic as of now is the Equilibrium. I'll take some time to investigate this application and come back with a response and some more changes in the text.

Noc2:
Just a quick reminder. It seems that not all contributors have signed the CLA.

semuelle (approved):
@DenisRomanovsky & @GreedyFerengi,
pending the signing of the CLA by all contributors to your PR and a description of your project in relation to Equilibrium, I am happy to give my approval.

DenisRomanovsky:
@semuelle , CLA is done. 
As of the Equilibrium, here is what we think about our economy models (added Aave and Compound too) :
Compound/Aave model - consists of manual liquidation where collateral+debt+fees go back to the protocol - nothing goes to the user-base, nothing to any other party.
Equilibrium's Bailsmen model - bailsmen are larger liquidity providers (essentially delegates for covering losses). liquidation is manual; debt and collateral passes to the 'bailsmen liquidity pool', which is a separate ring-fenced pool only for bailsmens' assets. Collateral+debt etc from said liquidation are shared amongst the pool's delegated bailsmen according to the relative % size of their stakes in the pool. This will later also involve weighting of the bailsmen in a later iteration of equilibrium. Nothing goes to the user-base. Bailsmen are backstopped by the Stability Fund in case of very large liquidations, or where the bailsmen's liquidity pool is insufficient to cover the entire liquidation call.
Minterest model - Liquidation is automated in the smart contract. Collateral and debt are collected via the contract and MOST of the initial liquidations income is kept in the liquidation pools as the liquidity in the various pools increases. When this tops out (likely pretty quickly), it will then be kept as part of the treasury surplus that buys MNT to grow the Minterest protocol.

alxs:
Hi @DenisRomanovsky, I'd be happy to support this too. At this point we're a bit sceptical of supporting "yet another lending protocol", but you seem a capable team & committed to the project, and you've pointed out some interesting differences between yours and other protocols. The only thing I would like to have a look at is the litepaper you mentioned, could you point me to it? I can't seem to find it.

DenisRomanovsky:
@alxs here is the [litepaper](https://docsend.com/view/k4mzyisngtgkead9/d/gn3utye9n7pgadgz)

alxs:
Sorry for the sudden change of opinion, but I'm not so sure about this any longer. If the only difference between you and Equilibrium is the liquidation process I don't see the need to support this, plus I don't get the impression that you need the funds.

I'm not very convinced by the litepaper and you promised a whitepaper over a month ago. Could you clarify how exactly you intend to support non-native assets without wrapping?

To be clear, I don't doubt the success of your project, I just don't necessarily think we should be funding this. Besides, the ETA for both milestones has already passed (at least you would have finished M1 long ago according to your roadmap). Could you comment on where you are with the development? I believe we should mostly be funding future work and not "buying back" work that has already been done.

All in all, this application has also been open for way too long. We'll give it two more weeks to gather the necessary approval and would otherwise close it unless you have a good argument against it.

mmagician:
After reading through your litepaper, I got confused by the notion of `automatic liquidation`. First, you say that this is what distinguishes you from other protocols - but then it looks like the liquidation action isn't automatic after all, but instead you shifted the responsibility from the "liquidators" to the treasury. If I understand correctly, the treasury still needs to perform a **manual** action of liquidating borrowers?
Perhaps I missed something - if that's the case, could you please detail the mechanism for automatic liquidation, as it's not something I found specification for in your document.

DenisRomanovsky:
@mmagician @alxs , agree, the protocol development went far beyond the milestones described in this application. We decided to close this application. But there is one piece of our work, which will be beneficial for Web3 community. I'll come up with another application, more detailed and better prepared, as soon as it is ready.

PR #219, opened 2021-01-23
 Implement raft consensus and collator

harodggg:
# Grant Application Checklist


mmagician (changes requested):
Thanks for the application, but your application lacks both high-level and technical details.
For example, I don't understand whether you are focusing on integrating an alternative consensus mechanism or working on a storage solution?
Furthermore, while I'm not an expert, my understanding is that Raft is not a BFT mechanism and albeit efficient, is best suited for permissioned networks, where you don't expect participation of malicious nodes. Could you elaborate on why a public blockchain might want to utilise Raft?

harodggg:
Hi, @mmagician.Thanks for your reply.First of all, I am sorry. What TiDot wants to do is to make collator have both storage functions and use raft consensus to store content in transactions. Collator only packages transactions, but these details are not what the grant will do. So I modified Tidot_network.md to focus on what we want to do in this grant instead of all the things TiDot wants to do. Of course, if you are interested, we can describe some of our relevant details to you.

harodggg:
Hi,The document has been updated with the addition of Current situation, Problem-solving ideas,Key part description and Detailed solution

mmagician:
Thank you for updating the application.

While your proposal is technically sound, the Research Team suggests that it is probably a little early to support this kind of work, at least until support for collator selection is developed.

What I mean by this is that no matter who the currently selected collator on the parachain is, other collators are still able to submit blocks to the relay chain validators, and these validators will continue checking blocks from "unauthorised" collators. The missing piece is for validators to be able to know who the "authorised" collator is, without necessarily checking the entire block - but rather only check the block headers, which should contain the collator's proof that they have the right to produce the parachain block (at a given block height of relay chain).

Furthermore, there are some reservations regarding the suitability of Raft over Aura, where constant block proposer rotation exists. The value proposition of Raft is high collator availability - if a leader fails, another one is promptly selected and block production doesn't stall. But it's a double-edged sword - the leader has control over inclusion of tx's into blocks as long as they stay online and active. This is a vector for censorship, which Aura avoids by ensuring rotation among nodes. That said, parachains have a choice of how they select block producers and I acknowledge there may be certain types of parachains which value availability over censorship-resistance. 

Considering these pros and cons, my suggestion would be to put this application on hold and revisit it once the development mentioned in the first point has been completed (or at least in beta).

mmagician:
@harodggg As mentioned in the previous comment, let's put this grant on hold for now and so I'll close the PR. Since it seems like your protocol could indeed find some use cases and you seem to have good knowledge of the substrate protocol, I strongly encourage you to re-apply in the future when the necessary collator-validator networking logic is in place.

PR #224, opened 2021-01-26
remove duplicate step & use updated get-changed-files action

mmagician:
- remove duplicate step with the same ID (one for local, one for real testing)

- Use the new version of file checking action: https://github.com/mmagician/get-changed-files/commit/1028587c8596c55a9d03d813a48aa1377f60b087. The change made there will allow squash & merge commits on PRs to function well, as linear commit history is not strictly required


Noc2:
Feel free to merge whenever you think it’s ready.

mmagician:
> Feel free to merge whenever you think it’s ready.

I hope it is. The thing is that while local testing is useful, it still won't let me try out merging a fork, for example. So things might still break unexpectedly, but it's going in the right direction :)

PR #235, opened 2021-02-01
PHP scale codec lib

gmajor-encrypt:
# Grant Application Checklist

### DIFF
--- a/applications/php-scale-lib.md
+++ b/applications/php-scale-lib.md
@@ -2,7 +2,7 @@
 
 * **Project Name:** PHP Scale Codec
 * **Team Name:** [gmajor](https://github.com/gmajor-encrypt) 
-* **Payment Address:** 3JJVnomZWTwUcb6tLAnESyy77NHi4cZ2hz
+* **Payment Address:** 0xC3094f0ddce699a1Ad9Ef2621DF68Cd297a4c44F
 
 
 ## Project Overview :page_facing_up: 
@@ -33,12 +33,12 @@ https://github.com/gmajor-encrypt/php-substrate-api
 ## Development Roadmap :nut_and_bolt: 
 
 * **Total Estimated Duration:** 8 weeks
-* **Total Costs:** 0.6 BTC
+* **Total Costs:** 12000 Dai
 
 ### Milestone 1
 
 * **Estimated Duration:** 4 weeks 
-* **Costs:** 0.3 BTC
+* **Costs:** 6000 Dai
 
 
 | Number | Deliverable | Specification | 
@@ -54,7 +54,7 @@ https://github.com/gmajor-encrypt/php-substrate-api
 ### Milestone 2 Example — Additional features
 
 * **Estimated Duration:** 2 weeks 
-* **Costs:** 0.3 BTC
+* **Costs:** 6000 Dai
 
 
 | Number | Deliverable | Specification | 
--- a/applications/php-scale-lib.md
+++ b/applications/php-scale-lib.md
@@ -1,0 +1,75 @@
+# Open Grant Proposal
+
+* **Project Name:** PHP Scale Codec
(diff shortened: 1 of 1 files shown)


gmajor-encrypt:
Thank you for your review, I saw it through https://github.com/w3f/General-Grants-Program/tree/master/rfp-proposal, I think it is more appropriate to mention proposal here

mmagician:
@gmajor-encrypt Thanks, yes that's the right place to open the PR. We will review it shortly

Noc2 (changes requested):
Thanks again for the application. Would it be possible to reduce the price? In general I recommend switching to a payment in DAI, since we will probably update the way we pay grant teams in BTC in the coming days. Also it seems you already previously worked on the project. Why did you do this? Are you potentially working on another project that would leverage this implementation?

gmajor-encrypt:
> Thanks again for the application. Would it be possible to reduce the price? In general I recommend switching to a payment in DAI, since we will probably update the way we pay grant teams in BTC in the coming days. Also it seems you already previously worked on the project. Why did you do this? Are you potentially working on another project that would leverage this implementation?

Thank you for your reply. I have replaced the BTC address with an Ethereum address. Yes, I did a part of the PHP scale codec work nearly 1 year ago, but we plan to change the PHP project to use Golang, and this part of the work has stopped. The Spring Festival is approaching recently, and I will have more time to complete the two open-source projects of scale PHP and substrate API

Swader:
I'd like to know more about the long term support commitment. What length are we talking here? Can you commit to keeping it up to date for 1 year since release? If so, I think the price is then fair.

(technically a codec should not care about types that update, it's just for decoding and encoding, but I think the bonus features are worth it)

gmajor-encrypt:
> I'd like to know more about the long term support commitment. What length are we talking here? Can you commit to keeping it up to date for 1 year since release? If so, I think the price is then fair.
> 
> (technically a codec should not care about types that update, it's just for decoding and encoding, but I think the bonus features are worth it)

@Swader hey, scale codec is a long-term maintenance task. Due to changes in the underlying substrate, runtime metadata and other basic data structures will change. If they are not updated in time, disasters will occur. I subscribed to the releases of the Polkadot library and updated it as soon as possible.

I guarantee that this will be maintained for a long time, not just for a year. In addition, I also have a substrate API PHP lib here. I understand that I cannot apply at the same time. I need to do this before applying for the substrate API PHP lib.

Noc2 (approved):
Thanks for the update. In general I’m willing to go ahead with it and will share it with the rest of the team.

Swader:
I agree with this and would like to proceed.
Maybe I am misunderstanding something though - SCALE is just a way to encode/decode, it has no reliance on types. It *can* be augmented with types (i.e. [scale-info](https://github.com/paritytech/scale-info)), but at its core, the encoding/decoding part should be standalone and long-term support-free just like BASE58 is.

So the way I see this is:

- PHP SCALE: standalone, works forever with basically no maintenance
- Additional types: probably via scale-info, maintained alongside it, useful as an "addon" for the PHP SCALE codec to be automatically compatible with Polkadot / Kusama (at that point it basically becomes PHP API for Substrate, but without the API connection).

This grant is mainly for funding 1), whereas the rest would belong more in the PHP API RFP.

gmajor-encrypt:
@Swader This is indeed the case. SCALE is a basis, but during the development process, the development problems I encountered were all that I needed to obtain the original value from RPC and parse it into a ``Human'' language. In short, DAPP developers were a little bit different. Concerned about the most basic types, and most concerned about what this storage/extrinsic/event is. Thanks again for your reply, I am ready to start tonight

Swader (approved):
@Noc2 let's merge this and get it going

mmagician (approved):

RouvenP (approved):

PR #237, opened 2021-02-01
Remote signing and slashing update 1

skardas:
# Grant Application Checklist

- [ ] The [application-template.md](https://github.com/w3f/Open-Grants-Program/blob/master/applications/application-template.md) has been copied, renamed ( "project_name.md") and updated.
- [ ] A BTC or Ethereum address for the payment of the milestones is provided inside the application.  
- [ ] The software of the project will be released under the Apache license version 2.0 as specified in the [terms and conditions](https://gist.github.com/Noc2/75bc58e8ce9b5d419ff883b0cf2b8c19).
- [ ] The total funding amount of the project is below USD $30k for initial grant applications and $100k for follow-up grants at the time of submission.
- [ ] The initial PR contains only one commit (squash if needed before submitting your PR).
- [ ] The grant will only be announced once we successfully [delivered](https://github.com/w3f/Grant-Milestone-Delivery) the first milestone.

### DIFF
--- a/applications/remote-signing-and-slashing-protection.md
+++ b/applications/remote-signing-and-slashing-protection.md
@@ -18,7 +18,7 @@ The validators play a crucial role on PoS networks due to operating stable, reli
 
 The proposed platform has independently separated the storage and the signing services from the validators and has provided a remote signing process through secure gRPC connections. This allows validators to create more robust and flexible operations, while providing additional security measures against possible attacks. Hence, it creates a more resilient solution compared to all other existing solutions, and in case of a compromise or uptime issues of the validator, Stkr can easily migrate to another validator without affecting users’ funds.
 
-The following diagram is used for ETH Staking Service that utilizes remote signing services
+The following architecture is used in Stkr.io for ETH Staking Service that utilizes remote signing services
 
 ![Overall Architecture](https://firebasestorage.googleapis.com/v0/b/test-sensor-veri.appspot.com/o/photo_2021-02-01_20-52-21.jpg?alt=media&token=d9c61b09-a5c2-4b04-8b82-3da2b09fed96)
 
--- a/applications/remote-signing-and-slashing-protection.md
+++ b/applications/remote-signing-and-slashing-protection.md
@@ -108,7 +108,7 @@ Slashing cases:
 ### Milestone 2 — Implementation for Remote Signing Services on Parachains (for both internal Testnet and Kusama)
 * **Estimated Duration:** 2 month
 * **FTE:**  4
-* **Costs:** 20000USD
+* **Costs:** 20000 USDT
 
 | Number | Deliverable | Specification |
 | ------------- | ------------- | ------------- |
--- a/applications/remote-signing-and-slashing-protection.md
+++ b/applications/remote-signing-and-slashing-protection.md
@@ -18,23 +18,23 @@ The validators play a crucial role on PoS networks due to operating stable, reli
 
 The proposed platform has independently separated the storage and the signing services from the validators and has provided a remote signing process through secure gRPC connections. This allows validators to create more robust and flexible operations, while providing additional security measures against possible attacks. Hence, it creates a more resilient solution compared to all other existing solutions, and in case of a compromise or uptime issues of the validator, Stkr can easily migrate to another validator without affecting users’ funds.
 
-The following architecture is used in Stkr.io for ETH Staking Service that utilizes remote signing services
+We have planned to implement the following architecture for Stkr's Polkadot staking service that utilizes remote signing services:
 
-![Overall Architecture](https://firebasestorage.googleapis.com/v0/b/test-sensor-veri.appspot.com/o/photo_2021-02-01_20-52-21.jpg?alt=media&token=d9c61b09-a5c2-4b04-8b82-3da2b09fed96)
+![Overall Architecture](https://firebasestorage.googleapis.com/v0/b/test-sensor-veri.appspot.com/o/Polkadot%20Design.png?alt=media&token=5ea6c795-52b1-48fd-aabe-22cda97d383e)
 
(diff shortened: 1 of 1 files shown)


mmagician:
Just as a reminder before reviewing any further: you have now removed the required deliverables 0a-0d from the [application template](https://github.com/w3f/Open-Grants-Program/blob/master/applications/application-template.md).
A tip: rather than submitting a new PR with updated application, you simply push more commits & your PR will be automatically updated.

mmagician:
> However, it should also be implemented in Polkadot codebase directly (i.e., not in Substrate).

Could you elaborate please? What changes in the Polkadot codebase would need to be done? Perhaps as a point of reference, you can have a look at [these developments](https://github.com/paritytech/substrate/pull/7628)

---

> This data will also be stored in a persistence database for future slashing prevention

Could you describe the exact logic for this? I assume there is a microservice running on the remote server which receives payloads and writes to database - but what's the mechanism for accepting/rejecting requests? Since this grant is for dealing with sensitive data, you'd need to describe your full approach before we can accept the grant.

---

Could you provide the architecture diagram for the Polkadot ecosystem, rather than eth2? While you can probably re-use some concepts, I imagine the entirety of the solution will look quite different.

---

> Providing a remote server

Let me repeat the question from the previous PR, since you haven't answered yet: we require that Grants Program teams to open source their code, so that users can, if they choose, run such a server themselves. Are you fine with doing this?

---
Finally, for an extra layer of security, have you considered using HSMs as part of the setup?

burdges:
I'd presume this simply deploys the somewhat functional remote signer in substrate now, yes?  We discourage much reliance upon remote signing until polkadot's key infrastructure stabilizes:

We need remote signing not to obstruct slashing reform back certs paritytech/substrate#7398 which achieves vastly more important slashing protections than remote signers do.  We'll deploy some fancy crypto ddpendent upon this for proving finality too, but still relatively remote signer friendly.  

We never properly designed a remote signer boundary for VRFs because VRF usages usually involve a call back.  We used VRFs suboptimally without a callback in BABE, which made BABE's remote VRF interface simple.  We'll fix this when Sassafras replaces BABE later this year.  Assignments VRF keys are being deployed within the next few months, which break VRF interface too, but they'll never require remote signing I guess.

We'll employ a lightweight remote signer friendly ring VRF for Sassafras on polkadot, which requires all block producers have equal-ish stake. We shall eventually provide a ring VRF option for parachains that want collators of unequal weight, but likely require collator remote signers produce SNARKs.

We'll have quite a different slashing story after slashing reform back certs and parachains launch: parachain validation keys become by far the most important component worth protecting, but doing so well requires validating entire parachains blocks on remote signers.  

All these features must be launched without anyone first providing remote signer support.

skardas:
@mmagician 
> Could you elaborate please? What changes in the Polkadot codebase would need to be done? Perhaps as a point of reference, you can have a look at these developments

The payloads for signing have been already implemented in Polkadot codease sources and Polkadot does sign_with method calls. We plan to modify substrate by adding additional sign_with_payload method then we will also modify polkadot sources as well.

>Could you describe the exact logic for this? I assume there is a microservice running on the remote server which receives payloads and writes to database - but what's the mechanism for accepting/rejecting requests? Since this grant is for dealing with sensitive data, you'd need to describe your full approach before we can accept the grant.

we propose to use remote signing services where all incoming data are checked against slashing status before being signed. In this respect, we need to provide the payload data that generates the signing data. Namely, for each signing request, the necessary metadata that generates the request should also be given. This data will also be stored in a persistence database for future slashing prevention. Before storing the data, a microservice checks that whether signing this data causes slashing or not. For example, (i) the health check for the whole epoch. (ii) Check that producing two or more blocks for the same slot. (iii) Check that signing multiple votes in the same round or on different chains. Although Polkadot nodes themselves have protection mechanisms against slashing, our proposal brings additional security protection against that kind of issue.

>Could you provide the architecture diagram for the Polkadot ecosystem, rather than eth2? While you can probably re-use some concepts, I imagine the entirety of the solution will look quite different.

new achitectural diagram is added

>Let me repeat the question from the previous PR, since you haven't answered yet: we require that Grants Program teams to open source their code, so that users can, if they choose, run such a server themselves. Are you fine with doing this?

we agree that all source code and libarary will be open source. In the delivery, we will provide open source licence

>Finally, for an extra layer of security, have you considered using HSMs as part of the setup?

For future plan, we may extend keymanager part to use HSMs in order for providing high level of security.

skardas:
@burdges 

>I'd presume this simply deploys the somewhat functional remote signer in substrate now, yes?

We are doing it now in substrate. We do not aim to provide the remote signer boundary for VRFs in Polkadot.

mmagician:
@skardas Thanks for elaborating on the proposal. As @burdges mentioned here and as we also discussed internally, it's probably not the best moment to develop a remote signing infrastructure, as the upcoming changes are likely to offer slashing protection baked into the protocol itself -> and as such are likely to render unusable your solution (in its current state) and potentially be harmful for validators using such setups.

In short, the mechanism proposed [here](https://github.com/paritytech/substrate/issues/7398) would mean that the keys which you wish to protect via a remote signer will no longer be relevant for slashing (or significantly less relevant).

We kindly ask you to familiarise yourselves with the above mentioned issue, as it's a prerequisite for further work involving signing. If after your research you still believe that remote signing could fit the proposed developments, I encourage you to submit another PR describing your idea in depth.

Alternatively, in case you already have extensive & proven experience in substrate & cryptography, we could support the development of https://github.com/paritytech/substrate/issues/7398 itself.

For now I'm closing this proposal and would like to thank you for all the effort put into it.

burdges:
We'll have more interesting keys worth protecting in future, after parachains launch and https://github.com/paritytech/substrate/issues/7398 lands, so hopefully less than six months.  

We'll still have churn after that but by then we'll be less paranoid about validators running setups that prove incompatible with us launching parachains.

PR #242, opened 2021-02-03
Update welcome.yml

Noc2:
Update Welcome message

### DIFF
--- a/.github/workflows/welcome.yml
+++ b/.github/workflows/welcome.yml
@@ -26,8 +26,8 @@ jobs:
         uses: thollander/actions-comment-pull-request@master
         with:
           message: > 
-            Congratulations! As part of the Open Grants Program, we want to help winning teams acknowledge their grants publicly. To that end, we’ve created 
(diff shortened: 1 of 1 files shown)


semuelle:
This bit is still too liberal in my opinion: `To that end, we’ve created a badge for grant-winning teams`. I would write `To that end, we’ve created a badge for projects that successfully delivered their first milestone`. `Projects`, so it's clear you don't get to use it in your next project.

alxs:
Agree with @semuelle, but I still think it would be cleaner to leave out the badge altogether and only let them know about it when they've delivered the first milestone (the announcement bit is still relevant, since they may want to get in contact prior to the delivery). We could set up an action that does the same thing as this one for first milestone deliveries.

I was helping @mmagician out with the GS integration over there, happy to look into this as well.

Noc2:
Integrated @semuelle suggestion. @alxs feel free to work on such a github action and update the message once you have integrated it. However I don’t think it’s an ultra high priority at the moment, since I think it’s overall pretty clear now and teams should be well aware of it.

PR #243, opened 2021-02-03
NewOmega Application

celrisen:
# Grant Application Checklist

- [X] The [application-template.md](https://github.com/w3f/Open-Grants-Program/blob/master/applications/application-template.md) has been copied, renamed ( "project_name.md") and updated.
- [X] A BTC or Ethereum address for the payment of the milestones is provided inside the application.  
- [X] The total funding amount of the project is below USD $30k for initial grant applications and $100k for follow-up grants at the time of submission.
- [X] The initial PR contains only one commit (squash if needed before submitting your PR).
- [X] The grant will only be announced once we successfully [delivered](https://github.com/w3f/Grant-Milestone-Delivery) the first milestone.

### DIFF
--- a/applications/newomega.md
+++ b/applications/newomega.md
@@ -108,7 +108,7 @@ Milestone 2 — Working prototype with signing transactions
 | 0b. | Documentation | We will provide both inline documentation of the code and a basic tutorial that explains how a user can participate in the game, interact with their wallets etc. |
 | 0c. | Testing Guide | The code will have unit-test coverage (min. 70%) to ensure functionality and robustness. In the guide we will describe how to run these tests |
 | 0d. | Article/Tutorial | We will write an article or tutorial that explains the work done as part of the grant. |
-| 1. | Integrate game client with polkadot | We will, via polkadot.js, integrate a wallet solution and allow playing the game by interacting with the smart contracts to defend, attack, view leaderboards, and progress in game (basic) via experience points gained through combat. |
+| 1. | Integrate game client with polkadot.js for wallet management and signing txs | We will, via polkadot.js, integrate a wallet solution and allow playing the game by interacting with the smart contracts to defend, attack, view leaderboards, and progress in game (basic) via experience points gained through combat. |
 
 
 
--- a/applications/newomega.md
+++ b/applications/newomega.md
@@ -1,0 +1,153 @@
+# Open Grant Proposal
+
+* **Project Name:** NewOmega
+* **Team Name:** NewOmega
+* **Payment Address:** DAI: 0xBd774017Ea2ae5B9dec156bA8C987d74eF249b30
+
+## Project Overview
+
+### Overview
+
+* Description
+
+NewOmega is, very simply, a blockchain game that can not be shut down.
+
+The current state of online gaming is a massive in-game purchases system, for anything ranging from quality of life improvements for players, to actual advantages, which usually can otherwise be obtainted but take time to get. Purchasing things, combining gaming and finance (at least, a real money wallet), became the norm. That is certainly the sentiment we're aiming to take advantage of.
+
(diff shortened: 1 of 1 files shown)


Noc2 (changes requested):
Thanks for the application. Sounds like an interesting project. Just for clarification you are referring to the following contracts her: https://github.com/WiktorStarczewski/newomega.trinity/blob/master/hardhat/contracts, correct? Maybe you can link to them in the deliveries. This makes it easier for us to evaluate the deliveries later. Also it seems like you are putting a lot of information on-chain, which might make the game crazy expensive. Have you thought about integrating an additional second layer  solution like for example https://www.celer.network/?

celrisen:
> Thanks for the application. Sounds like an interesting project. Just for clarification you are referring to the following contracts her: https://github.com/WiktorStarczewski/newomega.trinity/blob/master/hardhat/contracts, correct? Maybe you can link to them in the deliveries. This makes it easier for us to evaluate the deliveries later. Also it seems like you are putting a lot of information on-chain, which might make the game crazy expensive. Have you thought about integrating an additional second layer solution like for example https://www.celer.network/?

Hi David, great question. So what's currently implemented, in the "attacking" smart contract is storing game state and fight replays on the chain (state directly in storage, replays in event logs). That is obviously costly, but mostly due to the storing and processing of the replays - if I just calculate the outcome and store basic win/lose information, I can get the gas costs of that contract down to something that starts to be acceptable from game economics point of view. (And I could do something like regenerate the replay itself in the frontend from say, the pseudo-random seed that was used, right now a naive block stamp, but obviously could be Oracle, to eliminate the need for logs storage). Of course the definition of acceptable here perhaps does not extend to ETH Mainnet right now, thats a known problem, but there are (or almost are/about to be) networks out there that would make this calculation cheap enough for the game economics to make sense (and of course here's hoping Polkadot can improve on the current ETH design quite significantly). Not to mention additional optimisation rounds that the contracts would receive over time.

So to sum up, I feel like I'd like to explore more the L1 angle, but certainly keeping an open mind.

I've also updated the milestones with the GitHub links, as you requested.

Noc2 (changes requested):
Thanks for the quick response. Makes sense and I’m generally interested in supporting your application, but I think it’s currently a little bit expensive. Could you potentially remove milestone 3 for now, since we can always sign a follow-up grant once you delivered the first PoC. Furthermore, would it be possible to reduce the price for milestone 1 and 2 a little bit?  This way it would be easier to accept for us. And I guess part of the milestone 2 deliveries is the UI + the integration of the polkadot.js extension. “Integrate game client with polkadot” might currently be a little bit misleading, since you won’t integrate into polkadot (for example as a parachain).

celrisen:
> Thanks for the quick response. Makes sense and I’m generally interested in supporting your application, but I think it’s currently a little bit expensive. Could you potentially remove milestone 3 for now, since we can always sign a follow-up grant once you delivered the first PoC. Furthermore, would it be possible to reduce the price for milestone 1 and 2 a little bit? This way it would be easier to accept for us. And I guess part of the milestone 2 deliveries is the UI + the integration of the polkadot.js extension. “Integrate game client with polkadot” might currently be a little bit misleading, since you won’t integrate into polkadot (for example as a parachain).

Ok, makes sense, I'd certainly be interested in further development after the two milestones, if we decide we're a good fit. To that end, I've removed milestone 3 for now and tweaked the price of 1 and 2. Let me know what you think.

Noc2 (approved):
Thanks again for the update. I will share it with the rest of the team.

BenWhiteJam (approved):

Lederstrumpf:
Hi @celrisen, cool looking demo so far! After I authorize Ropsten payments, the interface remains on "Waiting for blockchain". Which aspects of your demo are supposed to work? Loading assets takes ages, but I can watch the demos under "training" and "logs", which is good enough for me at this stage of the PoC. Anyway - no need to fix this since it's an Ethereum demo ;)

Approved from me too; looking forward!

Lederstrumpf (approved):

PR #248, opened 2021-02-04
Zk Rollup Milestone2 Modifying

NoCtrlZ:
The substrate has ethereum compatibility because of frontier so it's not necessary to implement some milestone.
I modified some milestones.
Please confirm and merge.
Thank you!

### DIFF
--- a/applications/zk-rollups.md
+++ b/applications/zk-rollups.md
@@ -102,13 +102,12 @@ Our second step is to implement a ZK Rollup operator and a prover pallet. In ord
 
 | Number | Deliverable | Specification |
 | ------------- | ------------- | ------------- |
-| 1. | Implement Operator API Pallet | Runtime that run the sidechain environment |  
-| 2. | Implement Prover API Pallet | Pallet that allows service provider to compose own ZK Rollup network and interact with mainchain through RPC |  
-| 3. | Integration Test | Test sidechain and mainchain work together correctly on substrate-based chain |  
-| 4. | Prepare Dockerfile | Dockerfile that allows developer to setup the environment through docker |  
-| 5. | Write Tutorial | ZK Rollup tutorial that allows developer to run example commands |  
-| 6. | Documentation | Document which describes how to deposit, withdraw and send, be validator, and init and participate network on substrate-based chain |
-| 7. | Publish an article | After the Web3 Foundation's review, we will publish an article which highlights ZK Rollups on Polkadot. Potentially, we will integrate the article to the Polkadot's medium. |
+| 1. | Implement Zk Rollup CLI | CLI that deploy contracts to substrate-based chain and, run sidechain and prover |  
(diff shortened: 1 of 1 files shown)


mmagician:
@NoCtrlZ Before approving this PR, I would like to check the first milestone in more depth. I'm currently working on it and you should hear back from me in a couple of days.

alxs:
Hi @NoCtrlZ. This significantly reduces the amount of work for this milestone, which I think should be reflected in the price as well.

mmagician:
@NoCtrlZ @alxs Let's first close the first milestone before we proceed with this one.

mmagician:
As briefly discussed in the [evaluation notes for M1](https://github.com/w3f/Grant-Milestone-Delivery/pull/120), I think it doesn't make much sense to continue with the grant in its current form.
The original intention was to implement the zksync components as substrate pallets, which:
1. Clearly is a larger task than building a CLI as proposed in this amendment, as it involves re-writing the core of zksync's logic (operator, prover) to be substrate-runnable.
2. Given the outcome of M1 which showcases that re-using zksync's setup works, I agree with this PR's description that it's good enough (and preferable) to keep using a well-tested & proven architecture, rather than re-implementing it all yourself. Right now, I also don't see any great advantages of native substrate integration.

Since this amendment would completely change the scope of the original contract, I am closing it. 

If you wish to apply for a separate grant focusing on CLI and other improvements to the setup, you are welcome to do so. In such case, I'd please ask you to first submit an amendment to the original contract, removing milestones 2 & 3, just to keep things clean.

Thanks for your continued interest in the Polkadot ecosystem and congratulations again on completing the first milestone!

SotaWatanabe:
@mmagician I just came back from other work to ZK Rollups. I agree with you. In the current form, the scope is changed dramatically and it doesn't make sense. I will talk internally and get it back to you.

SotaWatanabe:
Hello @mmagician I am sorry for being late. During the implementation of the milestone1, we realized that implementing a ZK Rollups pallet was possible but it would be an over implementation. I am sorry. This was identified during the implementation of the milestone1. 

At this moment, our best proposal is to implement CLI instead of a pallet. This option reduces implementation costs and makes it easier for Substrate developers to use ZK Rollups. 

There are many missing pieces on ZK Rollups and it is originally designed to be implemented on Ethereum. For example, ZkSync (made by  Matter Labs) supports geth but it can’t be used for Substrate because some dependencies haven’t been supported for Substrate. Hence, we would like to make the missing peaces instead of making this complicated.

 I think the best way to do so it to terminate the milestone2 and resubmit another proposal separately. 
 
 It would be great if we could hear your opinion on this.

mmagician:
It's understandable that the requirements change as you dig deeper into the proposals, thanks for being open about it.
That sounds reasonable regarding M2 - I've prepared a [PR](https://github.com/w3f/Open-Grants-Program/pull/355) that removes M2 from the application. We will look forward to your new proposal.

PR #249, opened 2021-02-05
DotHacks Application

dothacklab:
# Grant Application Checklist

### DIFF
--- a/applications/dothacks.md
+++ b/applications/dothacks.md
@@ -83,10 +83,6 @@ None.
 
 ## Development Roadmap :nut_and_bolt: 
 ### Overview
-* **Estimated Duration:** 6 weeks
-* **FTE:**  12
-* **Costs:** $30,000 DAI
-
 We split our milestones into 4 stages.
 
 1. First we plan to implement a bug bounty platform, without dispute resolvings. That is, we consider all parties honest at this stage.
@@ -96,7 +92,7 @@ We split our milestones into 4 stages.
 
 In this milestone, we will focus on the first stage. Future stages will be left as follow-ups.
 
-### Milestone 1(for stage 1)
+### Milestone 1 (for stage 1)
 * **Estimated Duration:** 6 weeks
 * **FTE:**  12
 * **Costs:** $30,000 DAI
@@ -111,7 +107,7 @@ In this milestone, we will focus on the first stage. Future stages will be left 
 + Implementation of frontends (DApp UX).
 
 ## Future Plans
-The mission of Dothacks is to build a fair and secure cybersecurity platform, operating on a fully decentralized, proprietary protected public blockchain.
+The mission of Dothacks is to build a fair and secure cybersecurity platform, operating on a fully decentralized, privacy-protected public blockchain.
 
 As future plan, we hope to initiate a world-influential community that will attract global whitehats, cybersecurity agencies and vendors to join DotHacks. DotHacks will be fully governed by the community.
 
--- a/applications/dothacks.md
+++ b/applications/dothacks.md
@@ -1,0 +1,83 @@
+# DotHacks: a privacy-preserving and fair Security Response Platform on a public blockchain
+
+## Project Description
+### Overview (Introduction)
+Blockchain world has been suffering from vulnerabilities, e.g.,
+
+1. in centralized exchange: MtGox hack (https://en.wikipedia.org/wiki/Mt._Gox),
+2. in cryptography: double-spending in Bytecoin due to small subgroup attack (https://jonasnick.github.io/blog/2017/05/23/exploiting-low-order-generators-in-one-time-ring-signatures/),
+3. in smart contracts: The DAO hack (https://en.wikipedia.org/wiki/The_DAO_(organization)).
+
+Moreover, since 1) the market value involved is massive, 2) there is still lacking clearly defined or practical legal protections for cryptocurrency users, and 3) the techniques used in blockchains are still emerging and keeping evolving, which makes Blockchain security specialists become more desired, the security concerns of blockchain systems become significantly crucial.
+
+However, there is no current Security Response Center (SRC) existing in the Polkadot ecosystem to mitigate the loss caused by blockchain vulnerabilities.
+ 
+In this RFP, we propose DotHacks, a SRC platform built on a public blockchain, Polkadot in specific, to enable trustless SRC service. What make it different from conventional ones are: 1) we make use of Polkadot blockchain as the public ledger for persisting records; 2) we use consensus to govern the arbitrations; 3) and we use cryptography to preserve privacy and fairness; 4) we use tokenomics to incentivize the participants.
+ 
+As Polkadot is a project aiming at providing a feasible solution to bridge all other public blockchain projects, we believe that, with its design goal and promising progress, Polkadot is becoming the center of the blockchain world. Thus DotHacks can serve as a comprehensive SRC platform for blockchain projects and achieve its best potential.
+
+### Project Details
+Conventional SRC and crowdsourced bug-bounty platforms (e.g., hackerone, bugcrowd) suffer from limitations including:
+
++ insufficient privacy protection for white-hats
++ insufficient privacy protection for vulnerability reports
++ centralized arbitration mechanism
+
+These are also cases for some decentralized crowdsourced bug-bounty platforms (e.g., https://dvpnet.io/).
+ 
+DotHacks addresses these issues in the following ways:
+
++ White-hats can choose to submit reports anonymously.
++ DotHacks splits each report into shards and encrypt each shard to mitigate sensitive data leakage. Though a report is sharded, its logic correctness and consistency can still be verified, see section #Sharded report verification for details.
++ DotHacks uses State Channel Network (a layer 2 solution). State channel network enhances report submission privacy by privacy-protected peer-to-peer communication, and the submission can be verified by smart contract once exiting the state channel network and submitted on chain. The submission may comes with a dispute, and DotHacks uses arbitration committee to address disputes.
++ DotHacks uses randomness for an arbitration committee election to improve fairness.
++ On-chain reports are encrypted and each of them comes with a time-lock puzzle. Full vulnerability details will only be disclosed and be public after a set period. Vendors can also choose to keep the full report private.
+
+In specific, DotHacks works as follows:
+
+**Fig.1 Bug-bounty program workflow** (from a Whitehat's perspective):
+![](https://github.com/dothacks/dothacks.github.io/raw/main/assets/whitehat.png)
+
+**Fig.2 Bug-bounty program workflow** (from a Vendor's perspective)
+![](https://github.com/dothacks/dothacks.github.io/raw/main/assets/vendor.png)
+
+**Fig.3 Bug-bounty program workflow** (from an Arbitrator's perspective)
+![](https://github.com/dothacks/dothacks.github.io/raw/main/assets/arbitrator.png)
+
+#### Sharded report verification
+To verify a report while mitigating sensitive data leakage, DotHacks shards each report and encrypt them during submission. Common used stage conclusions among shards are extracted, to examine logic consistency and enable cross-shard verification.
+
+![](https://github.com/dothacks/dothacks.github.io/raw/main/assets/sharding.png)
+
+![](https://github.com/dothacks/dothacks.github.io/raw/main/assets/shard2.png)
+
+Arbitration committee are elected by randomness, to ensure fairness. Neither the whitehats nor vendors can predict who are arbitrators and thus cannot bribe them. Arbitrators can neither collude with each other to learn the full report.
+
+## Team & Team experience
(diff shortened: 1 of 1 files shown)


mmagician:
@dothacklab Thanks for the application. In general the idea of bug-bounties is very useful, though I'm pretty sure we can't leverage the already existing code. Are you aware of the bounty functionality in the treasury-pallet? See:
https://github.com/paritytech/substrate/pull/5715 & https://wiki.polkadot.network/docs/en/learn-treasury#bounties-spending
I understand that the scope of this is slightly different:
- only the Polkadot council can approve the bounty
- a single resolver can be awarded from one bounty (unlike possibly multiple white hat hackers in your proposal?)

If I'm not mistaken, the first point should be resolvable by including the treasury pallet in your parachain, which has its own council. This should work in a manner similar to that outlined in your proposal.
The second point is a bit more tricky under the current state of the treasury module - it would either require creating multiple bounties or by allowing the curator to choose only one best solution. 

A possible extension to the module - partial bounty payout to multiple addresses - could solve the second problem, if indeed it is that grave. I would rather argue it is not a major problem (as you can always launch a second bounty after receiving a first successful report). Anyway, I'll propose a partial multi-payout in the [substrate](https://github.com/paritytech/substrate/issues/) repo, let's see what kind of a response we get. If this turns out to be something desirable, then maybe we could propose a bounty on Kusama or Polkadot for it - and you would be of course welcome to apply!

 ---
Regarding your other points: ensuring report's privacy should be rather simply and doesn't require "sharding" as proposal in your solution. Instead, you can encrypt the report data with curator's Public Key, upload that to IPFS & off you go.

---
> DotHacks uses randomness for an arbitration committee election to improve fairness.

The assumption here is that the proposing committee is trusted by the white hacker and acts with the goal of getting its vulnerabilities fixed. Therefore, selecting a random arbitrator from among the committee is an ungrounded complication, in my opinion.

dothacklab:
Thank you for your reply.
>Only the Polkadot council can approve the bounty,thefirstpoint should be resolvable by including the treasury pallet in your parachain, which has its own council. This should work in a manner similar to that outlined in your proposal.

In the architecture of DotHacks cybersecurity response platform, its permissionless blockchain based bug-bounty system is operated automatically and independently.

1. Bug-bounty programisinitiated by Vendorson chain, then whitehatscanparticipateinthe programand submit vulnerabilityreports. The process does not require approvalsfromrole likePolkadotcouncil in treasury pallet.
2. When there is no dispute, whitehat communicates with the vendor point-to-point, and the vulnerability report is not exposed to any third party. Only when there is a dispute, the arbitration node will intervene.
3. The DotHacks council and the arbitration nodes are not the same role. The arbitrator node can be any organization or individual that has security capabilities and meets the requirements of the application.
4. The distribution of bounties is completed determined by vendors. And vendors will pay for their own vulnerabilities. The arbitrator nodes are only responsible for verification, evaluation, and voting. There will be a vulnerability evaluation standard jointly developed by the community for arbitration nodes to comply.

>Asingle resolver can be awardedfromone bounty(unlikepossiblymultiple white hat hackersinyourproposal?), the second point is a bit more tricky under the current state of the treasury module - it would either require creating multiple bounties or by allowing the curator to choose only one best solution.

In the architecture of DotHacks, Vendor will release a bug-bounty program on blockchain, with a clearly definition of what network assets are available for penetration testing. There are several significant differences between design of DotHacks and treasury pallet.

1. In our view, Treasury bounty is more like a fixed task, and its bounty will be paid when the task is completed. However, bug-bounty program for vulnerabilities is continuous. This is because the number and type of potential vulnerabilities are unknown to vendors.
2. Each program may contain a large number of completely different vulnerabilities from different network assets. Therefore, Vendors will pay different amounts of bounty to different whitehats.
2. The bug-bounty for security vulnerabilities is not a detachable task, and there will never be one best solution in vulnerability bounty program. The vendors will issue reward only according to value of each vulnerability submitted by the whitehats.

Regarding IPFS: Pure encryption does not have the ability to connect and verify different fragments of the vulnerability. However, fragmentation is necessary and IPFS is a choice.Inthe beginning, our team hope to prevent relying on a third-party service. But this is very flexible.



>The assumption here is that the proposing committee is trusted by the white hacker and acts with the goal of getting its vulnerabilities fixed.

Instead, we avoid such an assumption by introducing randomness. And because whitehats, vendors andarbitratorsdon't know who will be elected, they cannot collude: neitherwhitehats or vendors cannot bribe thearbitrators, nor arbitrators can collude to learn the full report detail.

In general, DotHacks is designed to be a cybersecurity response platform based on a permissionless public blockchain, to ensure that DotHacks' business operates in a completely transparent, fair, and privacy-protected manner.

mmagician:
I understand your viewpoint regarding the flexibility of the bug-bounty pallet vs the treasury pallet and see that there could indeed be some advantages.
Regarding your further points:

First, if you check the implementation of the bounty feature, you'll find that there is a role of a curators, which works just like your arbitrator. While this is a party that's trusted by the elector (in the case, the council), having multiple random arbitrators with domain-specific knowledge doesn't solve the problem, as someone trusted **still** needs to verify that they indeed have that kind of skillset (a very hard problem in itself).

Second:
> Encryption without fragmentation ("sharding") does not have the ability to connect and verify different fragments of the vulnerability report.

But why, and how would you verify a fragment of an **encrypted** report? The way I understand it, one needs to read a full report in order to assess its viability, no? Unless you have some powerful automated logic for checking reports (which, arguably, is still a very human task) **and** you are able to run that logic on partial, encrypted data - which requires some powerful homomorphic encryption schemes - this won't serve of much use, in my opinion.

Lastly:
> And because whitehats, vendors and arbitrators don't know who will be elected, they cannot collude

Vendors have little or no incentive to collude with anyone - otherwise they wouldn't post the bounty in the first place. Regarding the trust placed in arbitrators, see my previous points.

While I haven't worked on the implementation of the treasury pallet, my thought is that perhaps the best approach would be to try & extend it rather than building a separate new pallet (maintenance overhead). I'm looping in @rrtti who could perhaps share some of his thoughts.

rrtti:
Thanks @mmagician - @dothacklab do you want to have a call to talk more about this? I'd like to understand the complete mechanism.

dothacklab:
hi @rrtti ! yeah it would be great if we can schedule a call.
What time will it be convenient for you? Maybe we can discuss the time via email?
Are you using is raul at parity.io? Ours is dothacks.official at protonmail.com

rrtti:
Perfect, sending you an email now.

dothacklab:
First of all, thanks so much for pointing that we can make use of curator codes. This seems could indeed save us a lot of efforts. We would still like to keep randomness in curators election and have some qualification mechanism in election (we would explain later). But it's nice that we can extend and try to build on top of curators codes.

Regarding "vendors have little or no incentive to collude with anyone". In fact, they do. This happens during dispute resolving, when vendors and whitehats have different Severity Level ratings for a vulnerability. For example, a vendor thinks it's a low-impact one, but a whitehat thinks it's a severe one, a vendor may try to collude with arbitrators to get them rate it as low-impact, so that the vendor can pay fewer rewards. Vice versa, a whitehat may wanna collude with arbitrators to get them rate it as severe.

---

Regarding "an arbitrator may wanna collude with other arbitrators", this is the most interesting part, and this can answer both "trust placed in arbitrators" and "how a fragment of an encrypted report can be verified". In fact, the detailed workflow works as follows. Say a report is fragmented by a whitehat into 3 parts. He then also extracts a common stage conclusion A drawn from fragment 1, which is also used in fragment 2. For example, he gets the database password in the first fragment  part, but to conduct further attacks described in fragment 2, let's say,  getting operation system password, , he will need such a database password. If an arbitrator node knows such a stage conclusion, i.e., the database password, and the fragment 2, the arbitrator node then knows that the logic in  fragment 2 makes sense. If fragment 2 logic makes sense, and other people veriy that fragment 3 logic make sense, and all other fragments logic make sense, then it mean the full report logic makes sense.

Common stage conclusions and fragments need to be asymmetrically encrypted and need to be sent to corresponding arbitrator nodes, so that nodes receiving fragment 3 won't know fragment 2. This mechanism mitigate risks, because an arbitrator node still don't know the full vulnerability details,  but only a part of the vulnerability, and cannot run the attack on his own. 

These are all processed in layer 2. Once an arbitrator node, corresponding to his public key, is elected, a fragment and an associated stage conclusion is assigned and sent to him encrypted using an asymmetric ephemeral key. The fragment ephemeral key is derived from arbitrator node public key and report id, so only he can decrypt it. Message digests of the original text  of common stage conclusions and fragments also need to be broadcasted or be registered onto state channel network smart contract, so that arbitrator can check whether the whitehat is cheating during the dispute, for example, saying password is "111" as stage conclusion A in fragment 1, but using password is "222" as stage conclusion B in fragment 2; or sending different informations to arbitrators in a same group.

So to conclude, to start a dispute, he must fragment the report and extract some common stage conclusions; and we can see from the workflow that, if no encryption on the report, or arbitrators know each others, then arbitrators can learn the full report detail, and try to run the attack on their own.

---

Regarding "needs domain-specific knowledge".
Yes, at the begining, we might prefer to cooperate with other DID solutions in Polkadot ecosystem, and only elect from those certificated security experts. But in the future, as there is a ranking system in dothacks, whitehats get ranking scores by bug bounty programs, we can elect randomly from high-rank whitehats, while their identity can still be kept anonymous.

mmagician:
@dothacklab Thanks for explaining your proposal in greater depth. After discussing internally with @rrtti, we agreed that rather than creating a separate pallet, the substrate ecosystem would benefit more by extending the current treasury pallet to include some of your proposed features. The goal would be to merge your code upstream to the substrate repo's pallet-treasury upon successful completion. In particular we would you like to focus on:
1. Choosing a curator at random from a pre-agreed list of curators that applied for a specific bounty (needs some kind of verification of experience & Sybil resistance)
2. Submitting a private report, so as not to expose any security vulnerabilities to the public.

If this sounds like a good idea to you, let me propose two paths for pursuing this: continue via the Open Grants Program; or submit a proposal on-chain (and in the end apply via the existing bounty feature). It might make a bit more sense to apply on-chain, as the process would be transparent start-to-end for the substrate developers - and eventually they are the ones to accept your upstream PR.
For either of these paths, I recommend first getting hands-on experience & familiarise yourself extensively with the current state of pallet-treasury. Then, you would submit a detailed technical proposal for the two above features. 

Please let me know which would be your preferred option so we can help you throughout the process.

dothacklab:
Thanks for your advice. I agree that this approach is beneficial to the substrate ecosystem. And in our idea, DotHacks will run as an independent project, where pallet can be implemented by extending Treasury according to your requirements, so we will choose the first option and make this part of work as our millstone in the resubmission open grant application. We don’t know if this approach meets the goals you gave.  If it does, we will prepare the proposal. Looking forward to your reply.

rrtti:
If in the future you decide to apply for treasury funds, happy to walk you through the process.

mmagician:
@dothacklab I propose you start with simply extending the treasury pallet and try to get the new features merged upstream.
If that is successful, then we would be happy to support evolution of your DotHacks project - but I see these as two independent goals. If that makes sense, feel free to start working on the extension proposal.

dothacklab:
After reading the code related to the treasury pallet, we have few thoughts on a new solution.

The treasury module stores funds collected from transaction fees, slashing, etc, and it is managed by the council. We believe that the purpose of the treasury pallet is to make Council use the funds more effectively. So by extending treasury pallet, Council will be able to use these funds in aspects of cybersecurity response.

In our previous application, DotHacks is a public crowdsourced cybersecurity platform as Polkadot parachain, and the funds are deposited by external vendors participating in DotHacks's business. In other words, external multiple vendors will pay for their own cybersecurity services to the whitehats on DotHacks. It may not be consistent with the design intent of the treasury pallet if we add external vendors to make it public.

For the new plan, we will give priority to extend the treasury pallet by adding a new spending instrument named Bug-bounty. Bug-bounty instrument will be independent of the existing instrument of Proposal, Tipping, and Bounty. After implementing the Bug-bounty Instrument, the Polkadot Council, or any other Substrate blockchain Council, will be able to create a private cybersecurity response center respectively, by publishing bug-bounty programs, receiving vulnerability reports and issuing bug-bounty reward for its own assets through the treasury.

We will give more details about our plan in the new open grant application based on the above scheme.

mmagician:
>  In other words, external multiple vendors will pay for their own cybersecurity services to the whitehats on DotHacks. It may not be consistent with the design intent of the treasury pallet if we add external vendors to make it public.

You're absolutely right. I think it's best to think of these new features as a separate project for now.
That said, I believe that after modifying the treasury pallet with the bug-bounty feature, it will be relatively easy for you to take the developed code and still pursue the other direction of external vendors in the DotHacks idea.

dothacklab:
Sorry for the late response, our team were preparing for a conference in last few days. Now the new version of the application has been submitted. The link is as follows: https://github.com/w3f/Open-Grants-Program/pull/306
If there is any question please feel free to contact us.

PR #253, opened 2021-02-09
Denominating in usd

Noc2:
denominating in USD, mostly the following: The exchange rate for BTC depends on the time of payment. 

Also see https://github.com/w3f/Grant-Milestone-Delivery/pull/97


Noc2:
Merging it since Alex accepted it.

PR #285, opened 2021-02-27
fix headings

semuelle:


README.md
```diff
@@ -6,12 +6,12 @@
...
 **:loudspeaker: Important:** The Web3 Foundation _Open Grants_ Program focuses on technology grants up to **USD 30k for initial grants** and **USD 100k for follow-up grants**, which are tracked **transparently** on GitHub and disbursed in **Bitcoin or DAI**.
 
-**_This page is also available in [Chinese (中文)](./README_CN.md)._**
+:earth_asia: **_This page is also available in [Chinese (中文)](./README_CN.md)._**
 
 ---
 
-- [Introduction](#introduction)
-  - [:clipboard: About](#clipboard-about)
+- [:wave: Introduction](#wave-introduction)
```
alxs, line 13: I think :information_source: would also work well here

alxs (approved):

README.md
Noc2: We use “i” for the license, see https://github.com/w3f/General-Grants-Program#information_source-license. Maybe we should add this here as well. Although it doesn’t matter too much.

PR #291, opened 2021-03-02
Action to automatically label by approvals remaining

semuelle:
Should automatically label add and remove these labels when an approval is added.

image


Noc2:
Let's try it

PR #297, opened 2021-03-04
LevSwap- a decentralized  synthetic   leveraged token exchange

levswap:
# Grant Application Checklist

### DIFF
--- a/applications/EverlastingCash.md
+++ b/applications/EverlastingCash.md
@@ -1,0 +1,228 @@
+# Open Grant Proposal
+
+* **Project Name:** Everlasting Cash
+* **Team Name:** Cycan Technologies
+* **Payment Address:** ETH address: 0x722043b40dEb4656a103aF03Fd8E781d0146d3F6
+
+
+## Project Overview :page_facing_up: 
+
+### Overview
+  * While the incumbent stablecoins in the market have served the purpose of stabilizing the prices of cryptocurrencies, there are obvious limitations. Fiat-backed stablecoins such as USDT are not decentralized with the risk of being compromised by the central entity; while crypto-backed stablecoins are more vulnerable to price instability, since they are pegged to other cryptocurrencies. The risk of a possible crash for non-collateralized stablecoins is relatively high, since they lack collateral to liquidate the coin back into, while commodity-backed stablecoins might have liquidity problems as physical assets are not so easily liquidated. 
+  * Everlasting Cash is a new kind of stablecoin which solves these issues with a truly decentralized, anti-inflation and stable product with reserves.
+  * Everlasting Cash a.k.a. ELC is a decentralized algorithmic stablecoin using Substrate with a unique reserve system, i.e. a hybrid of a crypto-collateralized and an algorithmic stablecoin mechanism, with the collateralized mechanism providing the underlying value guarantee, and the algorithmic mechanism incentivizing the participants on one hand, and hedging the downside risk when the demand for the stablecoin is insufficient on the other.
+  * ELC aims to provide users in the Polkadot ecosystem with a stablecoin that appreciates naturally and achieves the goal of anti-inflation. 
+
+### Project Details 
+
+Everlasting Cash (ELC) contracts are written in Rust language and can be deployed to Polkadot/Kusama parachain (ink! pallet).
+
+#### Technical Design Overview
+
+<img src="https://github.com/CycanTech/docs/blob/main/Technical_Design_Overview.png" width="900">
+
+#### Tokens
+1) Everlasting Parachain Token - ELP
+ELP is a native token of the Everlasting Parachain (a parachain on Kusama). Participants are able to send ELP to the ELC multifunction contract as reserve assets to exchange ELC: according to the different system liability ratios, the contract will return both ELC and rELP, or rELP only to participants (see “Reserve” paragraph below).
+
+2) ELC Token
+ELC token is the stablecoin within the ELC contract system, with the initial pegged rate of 1 USD value in the year of 2021. The pegged rate varies over time based on inflation and depends on whether its supply expands or shrinks based on the system LR. 
+
+3) rELP Token
+rELP token is a risk asset in the ELC system (an equity in the system). Participants are able to add rELP liquidity to the multifunction pool, which could be realized by sending ELP to ELC contract. The system encourages user to do long-term liquidity mining so as to keep the reserve pool more stable with liquidity.
+
+#### Multifunction Pool
+1) Reserve
+Participants can send ELPs to the ELC system in exchange for both ELC and rELP, or rELP alone, depending on what the system liability ratios are. When the system LR is less than 30%, participants who send ELPs to the ELC system can obtain a certain amount of ELC and rELP minted and the LR of the system remains unchanged. When the system LR is higher than 30%, participants who send ELPs to the ELC system will only get rELP,  and this can decrease LR.
+
+2) Risk Reserve Fund
+The risk reserve fund includes two assets (ELPs and ELCs). The purpose of the Risk Reserve Fund is to keep the ELC prices floating between 98% and 102% of the ELC aim price (ELCaim). For example, 1 million ELPs out of the overall 21 million of ELPs initial issuance are used as initial risk reserves. 
+Such risk reserve fund is used as follows:  when ELP price is lower than 98% of ELCaim, the ELC system will swap ELP for ELC until all ELP in the pool are swapped into ELC to keep the price back to ELCaim. When ELP price is higher than 102% of ELCaim, the ELC system will swap ELC to ELP until all ELCs in the pool are swapped into ELP to keep the price back to ELCaim.
+
+3) Liquidity Mining
+Participants who have deposited ELPs to the reserve pool contract, can obtain rELP and use rELP to participate in the liquid mining. In order to maintain the size of reserves, participants are encouraged to do long-term liquidity mining. Participants who join liquidity mining will earn more tokens if they hold the tokens longer. 
+
+#### Inflation Governance
+The generation time of ELP is 6 seconds per block. The ELP target price rises every 10,000 blocks in K, K is an anti-inflation factor. The inflation factor K can be adjusted through the ELP governance mechanism. rELP holders can vote for the adjustment of the anti-inflation factor K when the USD inflation goes into hyperinflation.
+
+#### Oracle
+The oracle of reserve asset ELP and stablecoin ELC is implemented with Chainlink-Polkadot. We utilize Chainlink's decentralized Oracle network to quickly and securely connect end-to-end all the inputs and outputs of smart contracts to avoid pitfalls associated with deploying our own Oracle, such as long delays, additional charges, and even fatal security vulnerabilities. 
+
+#### Contract
+1) ELC - Minting and burning of the ELC token
+Mints ELC amount to the recipient account.
+Burns ELC amount from the account. 
+
+2) rELP - Minting and burning of the rELP token
+Mints rELP amount to the recipient account.
+Burns rELP amount from the account.
+
+#### Multifunction Pool
+
+Mints some rELC amount, in exchange for some ELP amount.
+Mints some ELC and rELP amounts, in exchange for some ELP amount.
+Returns the amount of ELC and rELP.
+
+ELCaim(after adjustment) = ELCaim(before adjustment) * (1+K)
+LR = Value/P(ELP) * Amount(ELP)
+
+Where:
+- Value is the value of the ELCs that have been issued
+- P isthe price of ELP
+- Amount is the number of ELP in reserve pool
+
+
+When LR is lower than 0.3, the ELC system mints amounts of ELCs and rELPs, in exchange for amount of ELPs.
+
+△Amount(rELP) = p(ELP) * △Amount(ELP) * (1-LR)/p(rELP)
+△Amount(ELC) = p(ELP) * △Amount(ELP) * LR
+
+When LR is higher than 0.3, the ELC system mints amount of rELPs only, in exchange for amount of ELPs.
+△Amount(rELP) = p(ELP) * △Amount(ELP) /p(rELP)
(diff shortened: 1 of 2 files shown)


Noc2 (changes requested):
Thanks for the application. I have one initial question: How is your solution different from [laminar](https://github.com/laminar-protocol/laminar-chain) or [Shadows Network](https://github.com/ShadowsNetwork/shadows)?

levswap:
> Thanks for the application. I have one initial question: How is your solution different from [laminar](https://github.com/laminar-protocol/laminar-chain) or [Shadows Network](https://github.com/ShadowsNetwork/shadows)?

Laminar and Shadow Network are committed to doing synthetic assets, and the main problems they pay attention to and solve are the price anchoring of the subject matter and synthetic assets and the liquidation of pledged assets. 
But LevSwap provide users with a novel decentralized synthetic leveraged token exchange that allows users to synthesize and trade leveraged tokens of any asset.The difference：
1.The synthetics asset represent long and short of the object, users do not need to provide collateral.
2.The long token pool and the short token pool is relevant.Dynamic adjustment will be executed according to different strength.
3. Liquidation of the synthetic assets is necessary for Laminar and Shadow Network cause by collateral design, while not for LevSwap.

Noc2 (changes requested):
Thanks for the quick response. Could you integrate this into the application? Also could you provide more technical details/specifications as part of the milestones. How is the dynamic fee calculated? To be honest, I have a hard time to understand how you want to archive this without any collateral (unless the fee is extremely high to basically create some kind of collateral). Is there already a similar solution out there that doesn’t use a collateral?

levswap:
> Thanks for the quick response. Could you integrate this into the application? Also could you provide more technical details/specifications as part of the milestones. How is the dynamic fee calculated? To be honest, I have a hard time to understand how you want to archive this without any collateral (unless the fee is extremely high to basically create some kind of collateral). Is there already a similar solution out there that doesn’t use a collateral?

Thanks for your response.We have updated our application.In the update, we added two parts to explain the design of leveraged token and dynamic fund fee.And the milestone has also been updated.
The answer for the question is as follows:
What needs to be clarified is that the bull and bear token's value in LevSwap is also supported by collateral assets. However the collateral in LevSwap is different from traditional collaterals with liquidation such as collaterals in Maker、AAVE、SNX. There's no liquidation in LevSwap.

When users buy bull/bear tokens in LevSwap, they use collateral asset(eg. DOT or Stable Coin). The collateral users input forms a collateral pool with two virtual parts supporting the bull and bear's value respectively.

Bull and bear tokens's prices are both based on the same collateral pool, therefore the bull and bear token holders form counter-parties, one side gains and the other loses.

There are two virtual parts of the pool supporting bull and bear tokens' price respectively. When the target price changes, the collateral in two virtual parts will change according to the leverage ratio, one goes up and the other goes down, therefore the bull and bear tokens' price changes  according to the leverage ratio as well.

levswap:
@Noc2 hi, do we need to provide any additional information?

Noc2 (changes requested):
Thanks for the additional information and clarification. I have two follow-up questions: for pricing you say you will use the “oracle machine of UniSwap”. How do you plan to do this with substrate? Also on your website it says under Partner: uma project and chainlink. How did you already partner with these teams? Also given that your github repository currently doesn’t contain a lot of information. Do you have an ethereum application that is already deployed?

levswap:
> Thanks for the additional information and clarification. I have two follow-up questions: for pricing you say you will use the “oracle machine of UniSwap”. How do you plan to do this with substrate? Also on your website it says under Partner: uma project and chainlink. How did you already partner with these teams? Also given that your github repository currently doesn’t contain a lot of information. Do you have an ethereum application that is already deployed?

@Noc2 Thank you for your reply. 
We will solve the pricing problem on substrate with an oracle on polka network, such as chainlink or HazelWord, which we have been contacting and discussing. 
Partner's description on our website is not accurate, we corrected the official website. 
We have discussed LevSwap with UMA several times ,and we have been in UMA's developer mining program whitelist.At the meantime, we are actively communicating with Chainlink about pricing on the polka network.  
We are now trying to implement our MVP based on EVM (not a  specific chain) , and have not deployed on any chain yet.  As polkadot/substrate offers features like high-performance and cross-chain information transaction,  substrate will be our first choice as an implementation of our formal product version.

levswap:
@Noc2 hi David, do we need to provide further materials or modification. Thanks!

Noc2 (changes requested):
Thanks. I will share your application with the rest of the team. In the meantime, I have some follow-up questions: Is it possible to see the code of your evm implementation? Did you also consider developing just a smart contract instead of a pallet (see https://stackoverflow.com/questions/56040779/when-should-i-build-a-substrate-runtime-module-versus-a-substrate-smart-contract)?

levswap:
Yes, this is our code base for LevSwap evm implementation: 
https://github.com/levswap/LevSwap-Contracts
(now we not completely ready for opensource yet, we have invite you to access our github repo. ) 

As for the choice of smart contract or pallet: 
Smart contract maybe a good choice as the first version of our implementation, it’s more light weight, while our vision is to build a decentralized cross-chain protocol for users to trade long or short on any assets. To achieve this, we are intend to complete a pallet implementation in the coming months for better flexibility, scalability and interoperability, and  engage in polkadot ecosystem.

levswap:
@Noc2 hello David, what should we do in this phase? If more details are needed we are pleased to offer, thanks!

Noc2:
I just looked at your code and it looks like you just copied: https://github.com/CryptoIcarus/SynLev. Also see https://synlev.medium.com/how-synlev-works-walking-through-the-beta-version-8d9f0154df74 I previously asked you multiple times, if you know an ethereum project that is doing something similar. This looks exactly like your project. Therefore I’m closing the application for now. Let me know if this is a misunderstanding and I should reopen the application.

levswap:
@Noc2 sorry David. Maybe we have mistaken your questions, so we missed to list similar solution for leveraged token.

Leveraged token is not a completely new product. in 2018 FTX introduced leveraged tokens based on its own perpetual contract. After that Binance, Huobi, MXC and other exchanges also launched their own leveraged tokens.

Synlev is a new decentralized leveraged tokens protocol. However, after studying the model of SynLev, we find that there are some defect in their model:
In SynLev, when there is a remarkable imbalance in long/short equity, the protocol will award/punish the weak side for a rebalance. 
But the model only conduct on the weak side to achieve the adjustment of long/short power. If there are not enough short sellers, then there is a penalty for selling Bear tokens and a reward for buying Bear tokens. 
In doing so, there are two problems: 
1) it can only play a regulatory role in buying and selling Bear tokens, and this is unfair to the holders of Bear tokens. 
2) it is only aimed at the Bear side and when Bear holders are exactly the weak party, and the rebalance mechanism is limited. 

Then we try to design a new model (as you can see in our proposal, we conducted our own deduction). 
How we solve these problems & our innovation lies in that:
1) through adjusting the market force between the long and short parties, charging the strong side and subsidizing the weak side, this is a long-short game in the overall market, not just the game of the holders of Bear or Bull, and can play a more regulating role. 
2) dynamic rates are charged according to time, not just at the time of buying and selling, as long as there is a transaction. 
It will be concluded that the fee should be charged through the product of the rate and time. 

2. In SynLev, when the market is out of balance, the loss of liquidity providers is huge. We think that it is irresponsible for SynLev to provide this function to users when it does not address the risk of huge losses of liquidity providers (in fact, it has already happened). 
In LevSwap, we disabled the option for users to provide liquidity, while users are encouraged to take positions through rate adjustments, and we are planning more product portfolios to enrich our liquidity. 

and as we have mentioned, the current EVM implementation is not our product version, we just want take it for fast model verification with enhancement on several aspects. But for sure open-source community helped and enlightened us and we will also contribute our final implementation.

We think leverage & leveraged token protocol is an important area in DeFi, we sincerely wish to contribute our thinking and cognition for a diverse cross-chain ecosystem.

Sorry for any misunderstanding and Looking forward to your reply.

simonhsj:
@Noc2 Hi David, do we need to provide more information or materials, thanks!

levswap:
@Noc2 Hi David, do we need to provide more information or materials, thanks!

alxs:
Hi @levswap. The more significant issue here is that you shared someone else's work as your own.

Apart from that, I found a couple of significant issues with your application. I'm happy to have a discussion on these here, but to be honest I think the chances that we reopen the application at this point are pretty low.

Firstly the first equation in your application is already incorrect, as you don't care about the slope of the change in price but about the relative change. It also does not solve to the next equation, but to `y = k x + c`.

Your proposed theoretical relationship between the prices of the represented asset and the leveraged asset _can_ be obtained by solving a relationship between the relative change of both assets, but it does not accurately represent a 3X leveraged position. In fact, you can easily check that a twofold increase in price in the underlying asset results in a 9-fold return on the leveraged position. The rate of return is also polynomial and the disparity between a 3X position and your theoretical price only grows with increasing change in price.

This basically means that the rest of your calculations are based on a wrong assumption, so I don't see this working in any possible way.

If we assume for the sake of it that `n_1 y_1 = n_2 y_2`, which certainly satisfies the zero-sum condition and is your intended equilibrium situation anyway, you would need to react ridiculously aggressively to changes in the price of the represented asset (polynomially on both sides) in order to move the number of positions to the intended range, which would likely result in funding fees so high that the theoretical price becomes irrelevant. At that point, you may have a cool money game, but nothing resembling 3X leveraged tokens.

alxs:
Hi @levswap. The more significant issue here is that you shared someone else's work as your own.

Apart from that, I found a number of substantial issues with your application. I'm happy to have a discussion on these here, but to be honest I think the chances that we reopen the application at this point are pretty low.

For starters, the first equation in your application is already incorrect, as you don't care about the slope of the change in price but about the relative change. It also does not solve to the next equation, but to `y = k x + c`.

Your proposed theoretical relationship between the prices of the represented asset and the leveraged asset _can_ be obtained by solving a relationship between the relative change of both assets, but as soon as you deviate from a constant leverage it does not accurately represent a 3X leveraged token. In fact, if the leverage ratio even minimally decreases (or increases for the short side), this would result in a significant loss for token holders, independently of the direction in which the price is actually moving.

These are some major issues hence my stance on rejecting your application also from a technical perspective, apart from the already mentioned mishap. Nevertheless, it was a cool problem to look at and I wish you the best of luck with your endeavour.

levswap:
Hi, @Thanks for replying and discussing the details here and We are also glad to discuss the details here and look forward to your more in-depth feedback.

1. Why we care about the relative change in price?

The first thing to understand is that what we want to design is a leveraged token protocol, as leveraged token, its price change is relative change to the target assets’ price with a specific leverage. For example, ETH-BULL as a leverage token for ETH with leverage k, when ETH’s price goes up 1%, then the price of ETH-BULL should goes up (k*1%). 

As a result, we use the following formula to analyze the leverad token system.

$$ 
\frac{dy}{y} = k \frac{dx}{x}
$$

With the formula you mentioned `y=kx+c`, if x goes up 1%, the y goes up 1% too, it's not a leveraged token system.

2. The mathematics of the formula 

$$
\frac{dy}{y} = k* \frac{dx}{x}
$$

$$
ln(y) = kln(x) + c
$$

$$
e^{ln(y)} = e^{k*ln(x)+c}
$$

$$
y = e^c * {e^{ln(x)}}^{k}
$$

$$
y = C*x^k
$$

3. Leveraged Token deviate from a constant leverage 

When the long and short side net value of the leveraged token are imbalanced, the leverage ratio will deviate from the target levearage ratio (eg. 3X). In my view, this is a normal trade-off between the market long/short power and the leverage, this is the same with the leveraged token on CEX(eg. FTX, Binance), and this is why the Leveraged Token on FTX need to be rebalanced every day. On Binance exchange, the leverage ratio always deivates from 3X without rebalance every day.

levswap:
@alxs 
Thanks for replying and discussing the details here and we are also glad to discuss the details here and look forward to your more in-depth feedback.

1. Why we care about the relative change in price?

The first thing to understand is that what we want to design is a leveraged token protocol, as leveraged token, its price change is relative change to the target assets’ price with a specific leverage. For example, ETH-BULL as a leverage token for ETH with leverage k, when ETH’s price goes up 1%, then the price of ETH-BULL should goes up (k*1%). 

As a result, we use the following formula to analyze the leverad token system.

$$ 
\frac{dy}{y} = k \frac{dx}{x}
$$

With the formula you mentioned `y=kx+c`, if x goes up 1%, the y goes up 1% too, it's not a leveraged token system.

2. The mathematics of the formula 

$$
\frac{dy}{y} = k* \frac{dx}{x}
$$

$$
ln(y) = kln(x) + c
$$

$$
e^{ln(y)} = e^{k*ln(x)+c}
$$

$$
y = e^c * {e^{ln(x)}}^{k}
$$

$$
y = C*x^k
$$

3. Leveraged Token deviate from a constant leverage 

When the long and short side net value of the leveraged token are imbalanced, the leverage ratio will deviate from the target levearage ratio (eg. 3X). In my view, this is a normal trade-off between the market long/short power and the leverage, this is the same with the leveraged token on CEX(eg. FTX, Binance), and this is why the Leveraged Token on FTX need to be rebalanced every day. On Binance exchange, the leverage ratio always deivates from 3X without rebalance every day.

alxs:
Assuming `y = C*x^k` means `y` is dominated by `k`. Again, if `k` decreases, `y` would drop dramatically, independently of whether `x` goes up or down.

levswap:
This formula is a theoretical analysis to show the price relationship between leveraged token and the target asset. As a 3X leveraged token, its price relationship is `y = c x^3`. 

In engineering,  the price and leverage ratio changes transaction by transaction rather than continuously occurring like a continuous curve. Therefore we should use the differential formula `dy/y = k* dx/x` to analyze the  actual price changes in levswap.

```
// value: k, y, x

UpdatePeice() {
	x_new = oracle.getPrice();
	
	delta = (x_new - x) / x;
	
	// update y	
	y = y + y* delta * k;

	// update k and x
}

```

alxs:
That makes more sense. In any case, this is a final rejection note due to the following reasons:
- You listed fake partnerships on your website.
- You provided others' work as your own.
- Said theoretical analysis is wrong since you use variable `k_1` and `k_2` as exponents.

Apologies for the tone but I want to be clear here.

PR #299, opened 2021-03-05
Create PredictDao.md

PredictDao:
# Grant Application Checklist

- [x] The [application-template.md](https://github.com/w3f/Open-Grants-Program/blob/master/applications/application-template.md) has been copied, renamed ( "project_name.md") and updated.
- [x] A BTC or Ethereum (DAI) address for the payment of the milestones is provided inside the application.  
- [x] The software of the project will be released under the Apache license version 2.0 as specified in the [terms and conditions](https://gist.github.com/Noc2/75bc58e8ce9b5d419ff883b0cf2b8c19).
- [x] The total funding amount of the project is below USD $30k for initial grant applications and $100k for follow-up grants.
- [x] The initial PR contains only one commit (squash if needed before submitting your PR).
- [x] The grant will only be announced once we've successfully [delivered](https://github.com/w3f/Grant-Milestone-Delivery) the first milestone.

PR #313, opened 2021-03-15
Create XPredictMarket.md

XPredictMarket:
# Grant Application Checklist

### DIFF
--- a/applications/XPredictMarket.md
+++ b/applications/XPredictMarket.md
@@ -1,0 +1,241 @@
+# W3F Open Grant Proposal
+
+* **Project Name:** X Predict Market
+* **Proposer:** Wenzhu liang
+* **Payment Address:** DAI：0x6e30f740A57a330de35D0ecE0ADF1eEc3D8C6c34
+
+
+## Project Overview :page_facing_up: 
+
+### Overview
+
+X Predict Market is a decentralized prediction market. The objective of X Predict Market is to enable users to participate in the prediction process in various ways by creating topics, discussing, predicting and approving the results. Users are also incentivized by these engaging activities. Beyond the prediction, users can actively interact and socialize with each other within the social system. It is a way to gather people's wisdom to conduct historical analysis and future exploration. It can also benefit current decision-making. Therefore, our team wants to contribute to such a way to condense the wisdom of the people. It will be developed based on Polkadot's substrate framework.
+
+### Project Details 
+
+As an evolving decentralized prediction market, X Predict Market is composed of an initial proposal pool and a formal prediction pool. In the initial proposal pool, users can create a variety of interesting topics for prediction. Meanwhile, other users can vote on the feasibility of the topic, and it will enter the formal prediction pool after the vote is casted.
+
+Each option will generate a token, of which the market making will be automatically conducted by the new Fixed Product Market Maker (referred to as FPMM). Users can make their own analysis and judgment on the potential results, and then obtain prediction tokens by exchanging assets. Users can receive governance token POT as rewards by creating prediction topics, participating in predictions, successfully predicted, and providing liquidity. POT can be used as collateral for becoming a governance node and to participate in community governance events such as on-chain voting.
+
+![main_flowchart](https://raw.githubusercontent.com/XPredictMarket/docs/master/images/main_flow_chart.png)
+
+![architecture](https://raw.githubusercontent.com/XPredictMarket/docs/master/images/architecture.png)
+
+#### Key Functions
+
+1. **Creation of The Prediction Market**
+
+    Everyone can initiate a prediction proposal in the initial proposal pool through X Predict Market. When creating the topic, users should set up the prediction type, topic, optional results, settlement currency, transaction fee ratio, prediction period, and the actual date when the result is revealed. Besides, any number of settlement tokens are required to add for initial liquidity, as well as a text describing the key points and merits of the proposal, the source of information, how the results will be validated and secured accurately and etc. It is to win the support and attract other users to join the prediction.
+
+2. **Voting in Initial Proposal Pool**
+
+    In the initial proposal pool, all users can browse and vote for any prediction proposals created by other users, which draw their interests. During the vote, users will need to verify the connected wallet addresses and hold a certain amount of governance tokens. If the number of ‘support’ votes exceeds a certain basic amount, and the support rate surpass 50%, the proposal is passed. If the proposal is finally passed, the users who voted in favor will get the governance tokens of the initial proposal vote as rewards. If it does not pass, the users who vote against it are getting rewards instead.
+
+3. **The Prediction Stage**
+
+    Proposals that have passed the initial voting will be displayed in the formal prediction pool. Users can choose any of the ongoing prediction topics to submit their opinions. When the topic is created, the creator sets the settlement currency for transaction at the very beginning, such as DOT. 
+    Users can either make predictions or add liquidity to the prediction topic. Users will get transaction fee by adding liquidity.
+
+    When making a prediction, a certain amount of settlement tokens is required in the wallet, and then the settlement token is exchanged for the prediction tokens. During the prediction period, users are free to sell prediction tokens to make profits or stop losses in a timely manner. When the prediction result is officially announced, the tokens which are successfully predicted will be exchanged for settlement tokens at a ratio of 1:1, otherwise the tokens that fail the prediction will lose value.
+
+    Referring to AMM, we set up a new type of fixed product automated market maker (FPMM), which will automatically adjust the value of prediction tokens, eliminating the need for in-depth support of order book, and the liquidity is greatly increased.
+
+    As for the specific process of FPMM? For example:
+
+    ① User JACK created a prediction market X1, 100DOT of basic liquidity are deposited, and the probability is set to 50:50;
+
+    ② The market automatically converts 100 DOT into 100 X1-Yes tokens and 100 X1-No tokens;
+
+    ③ JACK received 10 liquidity certificates X1-LP exclusively for the prediction market;
+
+    ④ Another user Tom purchases 25 DOT worth of X1-YES tokens from the market.
+
+    ⑤ First of all, assuming the transaction fee is 2%, Tom needs to pay a total of 25.5DOT.
+
+    Then, 0.5 DOT is deposited as transaction fee, and the other 25 USDC will be converted into 25 X1-YES tokens and 25 X1-NO tokens.
+
+    Then, because the user chose YES, 25 X1-YES tokens will be given to Tom, and another 25 X1-NO tokens will flow into the market. There will be 125 X1-NO tokens in the market. In the absence of new or reduced liquidity, the sum of product of the number of tokens of X1-YES and X1-NO remains a fixed value because of the automatic market-making mechanism of FPMM.
+
+    Therefore, the amount of X1-YES tokens in the market will change to 100*100/125=80, which is 20 less than the original, and these 20 will also be given to Tom. Therefore, the total number of X1-YES tokens obtained by Tom is: 25+20=45.
+
+    ⑥ When JACK retrieves liquidity using LP, he will retrieve 80 DOT and 45 X1-N0. When the final prediction result is NO, 45 X1-NO can be exchanged for 45 DOT, and JACK will harvest 125 DOT; Otherwise, JACK will only have 80 DOT left. 90% of the transaction fee will go to the liquidity provider, and 10% to the topic creator. Since JACK is both the topic creator and the only liquidity provider, he will get the full transaction fee of 0.5 DOT. When there are multiple liquidity providers, they are divided proportionally according to the amount of LPs they hold. Part of the fee is allocated to the topic creators to encourage users to discover interesting prediction proposals and introduce them to other users. The larger the pie, the more the liquidity provider will gain, so this is a total win-win mechanism.
+
+    ⑦ Assuming that the final prediction result is YES, prediction of TOM is successful. When he converts X1-NO to DOT, the platform will charge a 0.5% withdrawal fee: 45*0.5%=0.225DOT, of which 50% will be allocated to the correct prediction Node, 50% is used as the technology development cost of the platform.
+
+    Compared with AMM, this can tokenize the two predicted options and control the sum of their prices to 1. The more people who buy one of the options, the higher the price, but the price will never exceed 1. The traditional AMM is only suitable for direct transactions between 1 token and another token, and the price can theoretically increase indefinitely.
+
+    ![swap](https://raw.githubusercontent.com/XPredictMarket/docs/master/images/swap.jpg)
+
+4. **Prediction Completed**
+
+    When the prediction period set by the creator of the topic ends, the result is pending. In this process, any new prediction can no longer be proceeded while liquidity can be removed.
+
+5. **Result Uploaded On-chain**
+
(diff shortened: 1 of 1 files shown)


Noc2:
Just to double check: This application is related to your previous application: https://github.com/w3f/Open-Grants-Program/pull/299, correct?  Why did you set up a new github account?

Noc2 (changes requested):
Btw. did you consider to use smart contracts instead of runtime modules: https://stackoverflow.com/questions/56040779/when-should-i-build-a-substrate-runtime-module-versus-a-substrate-smart-contract and do you have any previous code, that you could share?

XPredictMarket:
> Just to double check: This application is related to your previous application: #299, correct? Why did you set up a new github account?

Yes, these two applications are related. We were not so sure of the name changing process so we created a new github account, of which the content is consistent with the former one except for the project name. Sorry if we have caused you any misunderstanding.

XPredictMarket:
> Btw. did you consider to use smart contracts instead of runtime modules: https://stackoverflow.com/questions/56040779/when-should-i-build-a-substrate-runtime-module-versus-a-substrate-smart-contract and do you have any previous code, that you could share?

The reason why we adopt runtime is that we want to implement cross-chain functionalities for various assets by ourselves in the future to meet the needs for diversified assets supported in our forecast process, so we need to use the chain to realize this goal. On the other hand, the scalability of chain is stronger, and we may have more new ideas to deliver them in the future, and we hope to retain more possibilities in terms of better scalability. We have submitted tokens-pallet code on github, and this part of the code is open-sourced. Other functions are under development.

Noc2:
Thanks. Sounds good to me. I will share the application with the rest of the team. I have one more question: I took a quick look at your [tokens implementation](https://github.com/XPredictMarket/X-Predict-Market-Runtime-Library/blob/master/tokens/src/lib.rs) and noticed that it is referring to XRC20. I think chainx originally used this. Are you related to [chainx](https://chainx.org/)?

XPredictMarket:
> Thanks. Sounds good to me. I will share the application with the rest of the team. I have one more question: I took a quick look at your [tokens implementation](https://github.com/XPredictMarket/X-Predict-Market-Runtime-Library/blob/master/tokens/src/lib.rs) and noticed that it is referring to XRC20. I think chainx originally used this. Are you in contact with [chainx](https://chainx.org/)?

Thanks for your reply! We have no reference to ChainX. Just to clarify: We decided to develop an asset management module by ourselves to enable the support of multiple assets in the future. And we did refer to some technical details of ERC20 standard because we have been doing development under the framework of Ethereum for years. Another reason is that there is an “X” in our project name, thus we put it “XRC20”. But considering that our major use case is to Predict, the team changed the name to PRC20.

Noc2 (approved):

alxs (approved):

mmagician:
@XPredictMarket In general I would like to support your project, but before proceeding I have a few questions regarding the resolution mechanism.

1. From your [technical overview](https://github.com/XPredictMarket/docs/blob/master/Technial.md):
 >  When the voting ends, we use Doublemap(pid, account, oid) to count the result and the Doublemap(pid, did, vote) to count the votes of the proposal

If I understood correctly, users themselves vote for the outcome of a market? How safe is that and how do you plan to prevent Sybil attacks?

2. But then you use Oracle-as-a-Service reference, to be fetched via an off-chain worker. First of all, this seems to contradict your earlier idea of users voting for the outcome. Which is correct?
3. Still regarding OaaS: when a user creates a new market, they need to decide on the resolution source, correct? I imagine that every source will have a different API, response format etc. How do you plan to do the integration of all these different APIs into the OCW? Presumably, the market creator needs to be enough technically apt to understand the source's format and code it into the OCW, right?

KillerJi:
> @XPredictMarket In general I would like to support your project, but before proceeding I have a few questions regarding the resolution mechanism.
> 
> 1. From your [technical overview](https://github.com/XPredictMarket/docs/blob/master/Technial.md):
> 
> > When the voting ends, we use Doublemap(pid, account, oid) to count the result and the Doublemap(pid, did, vote) to count the votes of the proposal
> 
> If I understood correctly, users themselves vote for the outcome of a market? How safe is that and how do you plan to prevent Sybil attacks?
> 
> 1. But then you use Oracle-as-a-Service reference, to be fetched via an off-chain worker. First of all, this seems to contradict your earlier idea of users voting for the outcome. Which is correct?
> 2. Still regarding OaaS: when a user creates a new market, they need to decide on the resolution source, correct? I imagine that every source will have a different API, response format etc. How do you plan to do the integration of all these different APIs into the OCW? Presumably, the market creator needs to be enough technically apt to understand the source's format and code it into the OCW, right?
Thank you Mmagician, we appreciate your reply !
Regarding the Question 1:
The phrase “When the voting ends, we use Doublemap(pid, account, oid) to count the result and the Doublemap(pid, did, vote) to count the votes of the proposal” means that we will count the number of tokens of each option using Doublemap(pid, account, oid) & Doublemap(pid, did, vote). As supplementary: during the transaction/trading process, users can purchase the token of his predicted option while at the same time, the amount of tokens held by each address can be obtained accurately, hence it’s able to prevent any possibility of Sybil attacks. The voting at this phase does not determine the final prediction result.
 
Regarding the Question 2
The final prediction result is determined by nodes, which will upload the results through the off-chain worker module. In the initial stage, a governance committee will be formed by the project team (as in the team) and community members, who will also act as a governance node that can upload the result with a tag ‘trustworthy’ to avoid Sybil attacks. Later, users who stake certain amounts of governance tokens can become nodes. The result supported by more than 50% of the nodes will be the candidate result which will be submitted to the test of a period of publicity. The publicity period was originally set for 3 days, which can be modified through community vote. Users can report the wrong nodes who upload results inconsistent with the reality. It is required to stake more than twice the staked amount of governance tokens of the reported node, if the reporting node can not raise that amount of tokens, it can initiate a report asking for the support of other users (by staking their tokens). Once the staking succeeded, the report is officially established and will be accepting the voting verdict by the whole network within 3 days.
When more than 50% of the votes support the reporter, the report succeeds, and vise versa. When the report succeeds, half of the staked tokens of the reported node will be given to the reporter, and the other half of the tokens will be distributed to the supporters of the reporting nodes according to the voting ratio, and vice versa.
A new round of publicity period begins after a successful report. During this period of time, users are also allowed to report, but the staked tokens required that the tokens to be staked for reporting will be doubled under this circumstance. The result of this period will become the final one if no one files a report in the publicity period.
Before the release of the governance result, the tokens participating in the governance will be locked up, also avoiding the Sybil attacks. 
Theoretically speaking, the result from reality is already a given fact before the verdict. Users always tend to vote for the correct option. If there are the whales collude to act evil, then they first need to hold the vast majority of governance tokens, which means a lot of costs, then if they jointly tamper with the results and cause errors, it also means that the project has been kidnapped and lost justice. Therefore, the governance currency of such a project will lose value, which will eventually cause them to lose money and lose long-term cash flow, which is not worthy. On the other hand, after a successful report, the staked governance tokens will be lost, which is also a loss. Therefore, from the perspective of direct and long-term interests, nodes will be biased towards submitting correct results, which ensures that the project can maintain the birth of a fair result.

Regarding the Question 3
①When creating a new market, users don’t need to care about the source of data, which is uploaded by the nodes who staked their governance tokens. Each node has their own trusted data source, which should not be provided by the market creator, thereby avoiding the suspicion of cheating. The market may be completely in the control of the market creator if it provides the data source.
②Regarding the OCW working mechanism : OCW obtains the final result by some sort of technical manner (for example: http request, database query request, Binary protocol request based on tcp, the sensor transmits data through the serial port and parallel port),  then we use unsigned transactions with signed payloads ( a way that can verify who sent the data without transaction fee, reducing greatly the cost of data uploading) to write the data into the pallet’s storage by calling pallet call function. Therefore, the OCW we are talking about is not limited to the OCW in the substrate. The OCW mentioned refers to an application that provides unsigned transactions with signed payloads. This will ensure that after the market is created, they will respond to the result data source in a timely manner, so that they can upload the corresponding data using OCW without changing the code of the substrate. In this way, the response format and source issues of different APIs can be avoided, and there is no need to integrate a specific API into OCW.
③Based on the ② theory, the market creator doesn’t need sufficient technology, nor does it need to be programmed into OCW. The essence of OCW is to use the result to initiate a transaction after obtaining the data. We use the unsigned transactions with signed payloads that saves massive amount of transaction fee. We do not limit the need to upload results by OCW, so the creator does not need to understand the source format in the technical aspect, and only needs to get a result, and ensure that this type of transaction is initiated normally, so it is not necessary to have OCW.

XPredictMarket:
> @XPredictMarket In general I would like to support your project, but before proceeding I have a few questions regarding the resolution mechanism.
> 
> 1. From your [technical overview](https://github.com/XPredictMarket/docs/blob/master/Technial.md):
> 
> > When the voting ends, we use Doublemap(pid, account, oid) to count the result and the Doublemap(pid, did, vote) to count the votes of the proposal
> 
> If I understood correctly, users themselves vote for the outcome of a market? How safe is that and how do you plan to prevent Sybil attacks?
> 
> 1. But then you use Oracle-as-a-Service reference, to be fetched via an off-chain worker. First of all, this seems to contradict your earlier idea of users voting for the outcome. Which is correct?
> 2. Still regarding OaaS: when a user creates a new market, they need to decide on the resolution source, correct? I imagine that every source will have a different API, response format etc. How do you plan to do the integration of all these different APIs into the OCW? Presumably, the market creator needs to be enough technically apt to understand the source's format and code it into the OCW, right?

Thank you Mmagician, we appreciate your reply !
Regarding the Question 1:
The phrase “When the voting ends, we use Doublemap(pid, account, oid) to count the result and the Doublemap(pid, did, vote) to count the votes of the proposal” means that we will count the number of tokens of each option using Doublemap(pid, account, oid) & Doublemap(pid, did, vote). As supplementary: during the transaction/trading process, users can purchase the token of his predicted option while at the same time, the amount of tokens held by each address can be obtained accurately, hence it’s able to prevent any possibility of Sybil attacks. The voting at this phase does not determine the final prediction result.
 
Regarding the Question 2
The final prediction result is determined by nodes, which will upload the results through the off-chain worker module. In the initial stage, a governance committee will be formed by the project team (as in the team) and community members, who will also act as a governance node that can upload the result with a tag ‘trustworthy’ to avoid Sybil attacks. Later, users who stake certain amounts of governance tokens can become nodes. The result supported by more than 50% of the nodes will be the candidate result which will be submitted to the test of a period of publicity. The publicity period was originally set for 3 days, which can be modified through community vote. Users can report the wrong nodes who upload results inconsistent with the reality. It is required to stake more than twice the staked amount of governance tokens of the reported node, if the reporting node can not raise that amount of tokens, it can initiate a report asking for the support of other users (by staking their tokens). Once the staking succeeded, the report is officially established and will be accepting the voting verdict by the whole network within 3 days.
When more than 50% of the votes support the reporter, the report succeeds, and vise versa. When the report succeeds, half of the staked tokens of the reported node will be given to the reporter, and the other half of the tokens will be distributed to the supporters of the reporting nodes according to the voting ratio, and vice versa.
A new round of publicity period begins after a successful report. During this period of time, users are also allowed to report, but the staked tokens required that the tokens to be staked for reporting will be doubled under this circumstance. The result of this period will become the final one if no one files a report in the publicity period.
Before the release of the governance result, the tokens participating in the governance will be locked up, also avoiding the Sybil attacks. 
Theoretically speaking, the result from reality is already a given fact before the verdict. Users always tend to vote for the correct option. If there are the whales collude to act evil, then they first need to hold the vast majority of governance tokens, which means a lot of costs, then if they jointly tamper with the results and cause errors, it also means that the project has been kidnapped and lost justice. Therefore, the governance currency of such a project will lose value, which will eventually cause them to lose money and lose long-term cash flow, which is not worthy. On the other hand, after a successful report, the staked governance tokens will be lost, which is also a loss. Therefore, from the perspective of direct and long-term interests, nodes will be biased towards submitting correct results, which ensures that the project can maintain the birth of a fair result.

Regarding the Question 3
①When creating a new market, users don’t need to care about the source of data, which is uploaded by the nodes who staked their governance tokens. Each node has their own trusted data source, which should not be provided by the market creator, thereby avoiding the suspicion of cheating. The market may be completely in the control of the market creator if it provides the data source.
②Regarding the OCW working mechanism : OCW obtains the final result by some sort of technical manner (for example: http request, database query request, Binary protocol request based on tcp, the sensor transmits data through the serial port and parallel port),  then we use unsigned transactions with signed payloads ( a way that can verify who sent the data without transaction fee, reducing greatly the cost of data uploading) to write the data into the pallet’s storage by calling pallet call function. Therefore, the OCW we are talking about is not limited to the OCW in the substrate. The OCW mentioned refers to an application that provides unsigned transactions with signed payloads. This will ensure that after the market is created, they will respond to the result data source in a timely manner, so that they can upload the corresponding data using OCW without changing the code of the substrate. In this way, the response format and source issues of different APIs can be avoided, and there is no need to integrate a specific API into OCW.
③Based on the ② theory, the market creator doesn’t need sufficient technology, nor does it need to be programmed into OCW. The essence of OCW is to use the result to initiate a transaction after obtaining the data. We use the unsigned transactions with signed payloads that saves massive amount of transaction fee. We do not limit the need to upload results by OCW, so the creator does not need to understand the source format in the technical aspect, and only needs to get a result, and ensure that this type of transaction is initiated normally, so it is not necessary to have OCW.

mmagician (approved):
> It is required to stake more than twice the staked amount of governance tokens of the reported node

This might be infeasible for most of the small players if they are up against a larger attacker - even when asking for the support of other users. If I'm not mistaken, this actually lowers the barrier for the attacker, allowing them to control the result with only 33% of the network (since the others would need at least 2x).

---
> Therefore, from the perspective of direct and long-term interests, nodes will be biased towards submitting correct results, which ensures that the project can maintain the birth of a fair result.

I see your point regarding the long term and I agree. Still, I think anyone who's trying to game the system could do so with only a short-term profit in mind. Attackers could sell the tokens they used to perform the attack.

---
Nevertheless, I am happy to give this a shot, I hope you can figure out the solution to the mentioned problems in the later stages of your project. Good luck!

mmagician (approved):
> It is required to stake more than twice the staked amount of governance tokens of the reported node

This might be infeasible for most of the small players if they are up against a larger attacker - even when asking for the support of other users. If I'm not mistaken, this actually lowers the barrier for the attacker, allowing them to control the result with only 33% of the network (since the others would need at least 2x).

---
> Therefore, from the perspective of direct and long-term interests, nodes will be biased towards submitting correct results, which ensures that the project can maintain the birth of a fair result.

I see your point regarding the long term and I agree. Still, I think anyone who's trying to game the system could do so with only a short-term profit in mind. Attackers could sell the tokens they used to perform the attack.

---
Nevertheless, I am happy to give this a shot, I hope you can figure out the solution to the mentioned problems in the later stages of your project. Good luck!

XPredictMarket:
> > It is required to stake more than twice the staked amount of governance tokens of the reported node
> 
> This might be infeasible for most of the small players if they are up against a larger attacker - even when asking for the support of other users. If I'm not mistaken, this actually lowers the barrier for the attacker, allowing them to control the result with only 33% of the network (since the others would need at least 2x).
> 
> > Therefore, from the perspective of direct and long-term interests, nodes will be biased towards submitting correct results, which ensures that the project can maintain the birth of a fair result.
> 
> I see your point regarding the long term and I agree. Still, I think anyone who's trying to game the system could do so with only a short-term profit in mind. Attackers could sell the tokens they used to perform the attack.
> 
> Nevertheless, I am happy to give this a shot, I hope you can figure out the solution to the mentioned problems in the later stages of your project. Good luck!

Thank you very much for your support, we will continue to optimize the project.

PR #323, opened 2021-03-22
discount Apron Network for ParityAsia hackathon award

mmagician:
As discussed with the team and with @jimmychu0807, due to the participation in hackathon we will discount any prizes won from the grant amount.


mmagician:
I guess [this](https://github.com/w3f/Open-Grants-Program/pull/116#issuecomment-804799782) means we can go ahead.

Noc2 (approved):

PR #325, opened 2021-03-22
discount Skyepass based on hackathon participation

mmagician:
# Grant Application Checklist

As discussed with the team and with @jimmychu0807, due to the participation in hackathon we will discount any prizes won from the grant amount.


mmagician:
The team in question is [OK with this](https://github.com/w3f/Open-Grants-Program/pull/212#issuecomment-804092193)

PR #333, opened 2021-03-26
Polkatax - legalizing your staking income

pmensik:
# Grant Application Checklist

### DIFF
--- a/applications/polkatax.md
+++ b/applications/polkatax.md
@@ -1,0 +1,122 @@
+# Open Grant Proposal
+
+- **Project:** Polkatax
+
+## Project Overview :page_facing_up:
+
+The only two things certain in life are death and taxes - this project would help you to deal with the latter one. The idea is to create an application that will provide user an easy way to generate a nice CSV or XLS file that can be included to his declaration of taxes and therefore legalize his staking income. 
+
+### Project Details
+
+ - General purpose application for generating staking income document
+ - Support multiple countries (English speaking and Czechia/Slovakia first)
+ - Use of Subscan to gather rewards data
+ - Use various exchanges API to determine the dollar price of the underlying asset at the time reward was received
+ - Use Central Banks API to convert the USD price to the supported country native currency
+ - Provide taxation guides and information for supported countries
+ - Initial support for Polkadot and Kusama, add more Substrate-based projects along the way
+ - Add fiat gateway for credit card payments
+
+### Use case
+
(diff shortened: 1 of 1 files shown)


mmagician:
@pmensik Thanks for the application. @jonasW3F has worked on something similar, perhaps he can provide some feedback.
Also, we can only support the development of open source software, so we cannot fund things like `buy www.polkatax.com domain`.

jonasW3F (changes requested):
Dear Polkadotters

Thank you very much for your proposal. I read through it and the things you are planning are quite familiar to me. I programmed a tool which gathers staking rewards from Polkadot / Kusama addresses in a specific timeframe and cross-references this with daily prices in almost all fiat currencies. At the end, the tool creates a .csv export which can easily be used to hand in with your taxform. The tool can be found here: https://github.com/w3f/staking-rewards-collector

It is open-source and free to use and in my opinion there is not much benefit at funding a similar project.

Best,
Jonas

alxs:
Apart from that, we currently don't award multiple grants in parallel and I believe this also applies to both programs in conjunction.

Since you have another open [application](https://github.com/w3f/General-Grants-Program/pull/416) for the General Grants Program and given Jonas' comments, I'm gonna go ahead and close this for now.

Thank you for your understanding.

alxs:
@pmensik another option would be for you to apply for this grant first (the UI part is interesting nevertheless, since Jonas' tool is currently purely CLI) since it's much smaller and then for the Subauction one. Additionally you could apply for the latter through Open Grants as a follow-up grant, which should be easier.

pmensik:
@alxs @jonasW3F Great feedback, I didn't know such a project already exists. Good point with the follow-up grant, however, regarding the time-frame, we are quite eager to start developing both projects (with Polkatax having a priority since it can be finished really soon) - especially given the fact that the backend is already there. I am just worried that we might be able to finish the project sooner than w3f approves the grant :-) So how to deal with such a situation?

alxs:
That's up to you.

However, since you're asking, given
- the significantly reduced scope of this grant discounting infrastructure and backend
- considering that you plan to monetise it,
- seeing as you plan to work on it regardless of whether the application is accepted, and
- considering this might lead to you experiencing some delays and/or uncertainty, since both an application and a milestone would need to be reviewed and accepted before you could apply again,

I would recommend you to consider carrying the costs for this yourselves and to focus on the Subauction application. You can start work on that project a bit later than the accepted date if necessary, in which case just let us know.

This is just my opinion though and you should go for whatever works best for you. Just let me know if you choose otherwise and I'll reopen this application.

pmensik:
@alxs Ok, those are really fair points so let's close it and let's see what we can do about it later on. Thank you

PR #357, opened 2021-04-09
Subspace

rg3l3dr:
# Grant Application Checklist


semuelle:
Solid proposal. Thank you for the submission. 

> we do not plan to launch Spartan as an actual production parachain or independent blockchain. Instead, this will only result in an initial test network for evaluation purposes.

Just to clarify: if all goes well, you are indeed planning to launch a para- _or_ independent blockchain?

rg3l3dr:
> Solid proposal. Thank you for the submission.
> 
> > we do not plan to launch Spartan as an actual production parachain or independent blockchain. Instead, this will only result in an initial test network for evaluation purposes.
> 
> Just to clarify: if all goes well, you are indeed planning to launch a para- _or_ independent blockchain?

Thanks for the quick feedback!

We are not planning to launch Spartan (proof-of-space) as a parachain. However, we will later extend `pallet_spartan` (perhaps with a follow-up grant) into `pallet_subspace` (proof-of-storage) and plan to launch Subspace as a parachain. 

Basically our proof-of-storage is based on a proof-of-replication, and any proof-of-replication must include an underlying proof-of-space and a proof-of-retreivability (i.e. a merkle proof to some accumulator). So Spartan can be combined with a proof-of-retrievability to be a full proof-of-storage or proof-of-replication.

mmagician (approved):

Noc2 (approved):
This looks good to me. Potentially you could also explorer to use Proof-of-Capacity for [Collators](https://wiki.polkadot.network/docs/en/learn-collator#docsNav).

alxs (approved):
Looks good to me too. Keep up the good work and good luck applying to the Builders Program.

burdges:
As a rule, proof-of-work-like protocols like proof-of-space are incompatible with parahain networking on a fairly fundamental level. 

We cannot reliably support either XCMP messages or defend against parachain capture for proof-of-work or proof-of-space chains.  We either allow DoS attacks on our validators, or else we impose another proof-of-stake-like scheme upon them, both of which waste our developer resources.  We should therefore avoid encouraging work on straight proof-of-work-like protocols, including chia-style wasteful proof-of-space. 

As I understand it, this project includes enough proof-of-stake so that they could still receive XCMP messages and avoid parachain capture.  I also like proof-of-useful-work/space ideas, but..

We cannot make proof-of-work-like protocols run full speed, so either they spend a parachain slot suboptimally or else consign themselves to parathread status.  We cannot support mempool-free operation for proof-of-work-like protocols, so they inherently squander the CPU savings achieved by pre-batching transactions.  We want nodes to distribute data too, not merely possess it.

I designed a [proof-of-useful-work using mixnet cover traffic](https://github.com/w3f/messaging/blob/master/incentives/incentives.tex) way back in Sep 2018 before polkadot ate all our time, which I think helps addresses these concerns.  In brief, we'd perform work directed by randomness that's provided by VRFs, either sending cover traffic or else storage fetches, but then later select which work we reward also using VRFs, so that block production runs entirely separately.  It's worth chatting about these ideas sometime. 

We should chat about erasure coding too.  We've quite a speedy Reed-Solomon crate for Polkadot, with which our backing validators commit to parachain candidate blocks existence.  We only retain candidates for 24 hours in this state, but parachains could retain blocks longer if they used erasure coding too.  We'll eventually save parachain state check points using Reed-Solomon too, prsumably across validators but maybe across parachain nodes too.  We already avoid syncing full chains using a basic warp sync, so we should eventually sync by reconstructing from state checkpoints.  

I've designed a tit-for-tat incentive scheme suitable for when polkadot validators reconstruct parachain blocks, but does not exactly cover the chain sync story.  We'd therefore maybe find roundly this sampling approach useful for state checkpoints, perhaps on both the relay chain and on parachains. 

Although Reed-Solomn fits proof-of-stake models well, one might discover that other erasure codes fit other threat models better, but that's another topic.

PR #341, opened 2021-04-01
NFT Collectibles Wallet

GamePowerDev:
# Grant Application Checklist

- [x] The [application-template.md](https://github.com/w3f/Open-Grants-Program/blob/master/applications/application-template.md) has been copied, renamed ( "project_name.md") and updated.
- [x] A BTC or Ethereum (DAI) address for the payment of the milestones is provided inside the application.  
- [x] The software of the project will be released under the Apache license version 2.0 as specified in the [terms and conditions](https://gist.github.com/Noc2/75bc58e8ce9b5d419ff883b0cf2b8c19).
- [x] The total funding amount of the project is below USD $30k for initial grant applications and $100k for follow-up grants.
- [x] The initial PR contains only one commit (squash if needed before submitting your PR).
- [x] The grant will only be announced once we've successfully [delivered](https://github.com/w3f/Grant-Milestone-Delivery) the first milestone.

PR #418, opened 2021-05-17
Milestone update for the GamePower nft-collectibles-wallet

GamePowerDev:
This milestone update is for accepted application #341 

After much development has been done on our original Milestone 1: "Build NFT Collectibles Wallet Mobile App", It was realized that in order to make our vision of a NFT wallet that can work across any substrate based blockchain, we will need to implement a pallet that sits on top of the OMRL NFT library.

We originally had this as part of our #341 submission but removed it. However its inclusion would benefit this project greatly. Instead of creating an entirely new NFT pallet/standard, we are using ORML's library for NFTs. Any substrate chain that includes this pallet into their runtime will provide an interface for the wallet to communicate with. A list of methods that the pallet will provide has been added to our application for review. 

Since we have already started working on this pallet, we will not be asking Web3 to fund this milestone, however we would like it included in our application since it was started after our application acceptance.


semuelle:
Hi Michael, thanks for the update. Do I understand correctly that you are _extending_ the ORML NFT pallet and that any parachain with that pallet will still be compatible?

GamePowerDev:
> Hi Michael, thanks for the update. Do I understand correctly that you are _extending_ the ORML NFT pallet and that any parachain with that pallet will still be compatible?

Hello Sabastian! Yes, in a way we are extending ORML NFT pallet. We will be providing a pallet that gives a loose coupling to the ORML NFT pallet. There are only 2 methods that we will interface with on the ORML NFT pallet and that's "transfer" and "burn".

The reason we are making this a loose coupling is because any chain integrating with the wallet can choose to not allow burning of NFTs or transferring of NFTs on their chain through the wallet. A runtime can easily pass a bool to our pallet Config stating that there should be no transfers or burning... likewise if they allow transfers and burns but would like to perform domain specific logic before those methods are sent to ORML, they can pass a function to the "OnTransfer" and "OnBurn" Config params to handle that logic.

We are providing 4 methods in our pallet that will not interface with ORML at all and these are specific for wallet functions:
```
// list an NFT for sale
fn list(origin, asset_id: u64, price: T::Balance) -&gt; Result;
// buy a listed NFT
fn buy(origin, asset_id: u64, list_price: T::Balance) -&gt; Result;
// add an emote to the NFT (for social)
fn emote(origin, asset_id: u64, emote: Vec&lt;u8&gt;) -&gt; Result;
// allows a user to claim a minted NFT
fn claim(origin, asset_id: u64) -&gt; Result;
```

Our goal is to allow any chain that is running ORML to behave just as it does today. The NFT Wallet pallet will just live along side the ORML NFT pallet and piggyback off of its storage when displaying, transferring or burning a user's NFTs. In the future we may expand the wallet to be fully decoupled from ORML and allow chains to use whatever underlying NFT system of their choosing. But right now we really like the way ORML is handling NFTs and will stick with that for a while.

alxs:
Hey again @GamePowerDev. Thanks for the clear explanation, this makes sense and it looks good to me.

I have something to add not directly related to the additional milestone: note that deliverable `0d` is usually a Medium article that may be catered to less technical users, and it would be perfectly fine for you to only deliver it only with the last milestone. We would also cross-promote such an article. A tutorial as you describe it is usually delivered in the README and would already be part of the documentation (see the description of deliverable `ob`). So feel free to move what you describe as deliverable `0d` to `0b` and remove it wherever you don't intend to publish anything.

GamePowerDev:
Thanks for the feedback Aleixo! I've gone ahead and remove 0d from both milestone 1 and 2 because it only makes sense to write that article once at the very end (milestone 3).

alxs (approved):
Sounds good. I'll share with the rest of the team.

semuelle (approved):

Noc2 (approved):

PR #420, opened 2021-05-21
Klevoya WASM Smart Contract Fuzzer grant application

mtabz:
# Grant Application Checklist

### DIFF
--- a/applications/klevoya_fuzzer.md
+++ b/applications/klevoya_fuzzer.md
@@ -14,7 +14,11 @@ Currently blockchain developers mainly check their smart contracts using manuall
 
 Klevoya’s mission is to build automated solutions that help blockchain developers find and eliminate bugs and zero-day vulnerabilities in their code before hackers can get to them.
 
-**_Our solution:_** develop a WebAssembly byte-code fuzzer (i.e. dynamic execution of a software program in response to random inputs) for smart contracts that are to be deployed/executed in [Substrate’s Contracts pallet](https://substrate.dev/docs/en/knowledgebase/smart-contracts/contracts-pallet). This project builds on Klevoya’s previous work developing byte-code fuzzers for other WASM based smart contract chains.
+**_Our solution:_** The [Substrate Contracts pallet](https://substrate.dev/docs/en/knowledgebase/smart-contracts/contracts-pallet) will be used to execute smart contracts written by 3rd party developers, and these smart contracts may contain a variety of bugs. In this grant we will develop a WebAssembly smart contract byte-code fuzzer (i.e. dynamic execution of the smart contract in response to random inputs) for smart contracts that are to be deployed/executed in [Substrate’s Contracts pallet](https://substrate.dev/docs/en/knowledgebase/smart-contracts/contracts-pallet). We will not be fuzzing the Contracts pallet itself, only the smart contacts that are to be deployed on it.
+
+From experience with other blockchain ecosystems we have observed that the majority (if not all) of smart contract hacks that take place are a result of logic bugs. Therefore the main objective of this smart contract fuzzer will be to identify smart contract logic bugs. An ancilliary benefit will be to identify memory corruption errors in cases where developers use non-Rust programming languages (e.g. [Ask!](https://github.com/patractlabs/ask) which is based on AssemblyScript) to write their smart contracts.
+
+This project builds on Klevoya’s previous work developing byte-code fuzzers for other WASM based smart contract chains.
 
 **_Benefits of fuzzers:_**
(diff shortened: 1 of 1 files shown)


alxs:
Thanks for the application @mtabz. We'll look into it as soon as possible.

Noc2 (changes requested):
Thanks for the application. You should potentially also take a look at the following grant project: https://github.com/w3f/Open-Grants-Program/pull/9 or here: https://github.com/pventuzelo/wasm_runtimes_fuzzing. However, this grant was for  wasmi/wasmtime and not a polkadot or contract runtime. Why do you think fuzzing a runtime will be useful? Isn’t it possible to ensure through the statistical guarantees of rust that the runtime won’t crash.  Did you discuss this with parity directly, since they [developed the module](https://github.com/paritytech/substrate/tree/master/frame/contracts)?

Noc2:
Let me also ping @athei here, who is working on the contract module.

mtabz:
Hello @Noc2, yes we have seen the project https://github.com/w3f/Open-Grants-Program/pull/9, however as you have mentioned that project was focused on fuzzing wasmi and our goal is different.

> Why do you think fuzzing a runtime will be useful?

We're not planning on fuzzing the runtime in this grant. Rather our focus is on fuzzing the smart contracts that will be deployed on the contracts pallet. 

You are right that in general with Rust as a smart contract programming language there should be fewer memory safety issues found through fuzzing the smart contracts. However, there are other smart contract programming languages in development by other teams that do not offer the same compile-time memory safety guarantees (e.g. https://github.com/patractlabs/ask which is AssemblyScript based). In addition we also plan to use fuzzing to identify smart contract logic bugs. In this way we hope to help smart contract developers identify a wide range of security issues in their smart contracts.

> Did you discuss this with parity directly, since they developed the module?

We have not discussed this with parity. We'd be happy to do so if you can suggest a suitable person to contact?

athei:
As the de-facto maintainer of the contracts pallet I would welcome the addition of a fuzzer. However, after reading the proposal I still have one open question:

Given I understand your proposal correctly you want to fuzz the `pallet-contracts`( i.e find bugs in this piece of software). However you propose to not fuzz against the real software but your own implementation in order to reach an execution speedup (transpiling to C++). How are you finding bugs in `pallet-contracts` when you execute your input against another software?

mtabz:
Hi @athei, thanks for the response.

It appears that the grant proposal could benefit from some additional clarification around the intention as this seems to be unclear.

To clarify:

1. Fuzz target: The proposal is not to fuzz pallet-contracts. We assume that this piece of software is largely bug-free. However (and please do correct us if we are mistaken), the contracts pallet will be used to execute smart contracts written by 3rd party developers, and these smart contracts may contain a variety of bugs. Our intention is to fuzz those smart contracts, which is why we believe we can execute the smart contracts against our own model of the contracts pallet.
2. Logic bugs are the primary goal: yes we completely agree with you that most bugs will be logic bugs. We can see from other blockchain ecosystems that the majority (all?) of smart contract hacks that take place are a result of logic bugs. This is why the fuzzer we are developing will not only detect crashes; in the second half of the grant we plan to implement techniques that allow the fuzzer to detect common smart contract logic bugs.
3. Memory corruption errors are a secondary goal: as I noted in my [response above](https://github.com/w3f/Open-Grants-Program/pull/420#issuecomment-846143524) to David, ink! (and by extension Rust) is not the only language that smart contracts may be written in in the future, as we start to see other smart contract languages being developed by the community (like [Ask!](https://github.com/patractlabs/ask)) that are not based on Rust. Smart contracts derived from these non-Rust languages may exhibit memory corruption errors that a fuzzer may catch.

mtabz:
@athei we have updated the proposal with your feedback on making the project goal more explicit and included details of similar work. Let us know if there are still items we should add to make it clearer.

With regards to future maintenance, the plan is to have API level compatibility with the contracts pallet and underlying logic being mocked as much as possible. In this way we hope to reduce the maintenance to a minimum.

We are not overly familiar with sp_sandbox, but agree that we can perhaps investigate this in a follow on grant.

The latter two points have also been added to the grant proposal.

Noc2 (approved):
Thanks for updating this. Also thanks a lot @athei for helping review the application! I’m happy to go ahead with it.

alxs (approved):
Very happy to support this as well. Good luck with the development!

semuelle (approved):

alxs:
@mtabz works, no worries and thank you for the update! :)

alxs:
Hey @mtabz, any news?

alxs:
@mtabz note that since we haven't heard from you in a while and given the delay, if we don't hear from you in the next two weeks we'll assume you're no longer interested and terminate the grant.

PR #424, opened 2021-05-24
UpLink proposal

andre-francisco:
# Grant Application Checklist

### DIFF
--- a/applications/uplink.md
+++ b/applications/uplink.md
@@ -1,0 +1,294 @@
+# W3F Open Grant Proposal
+
+* **Project Name:** UpLink
+* **Team Name:** HypeLabs Inc.
+* **Payment Address:** 3Fwmy2QHw1Xp7S956PMG9NKq1yngzRqGqf (BTC)
+
+## Project Overview :page_facing_up:
+
+### Overview
+
+UpLink is a project that proposes a decentralized and infrastructure-free approach to peer-to-peer connectivity. The system establishes local mesh networks between devices, enabling them to connect directly, cooperate, and relay content on behalf of each other. By cooperating, the devices spend resources, such as bandwidth, memory, and computational power. The consumption of the device’s surplus resources, however, creates an incentive for users to not cooperate, effectively hampering the network and reducing its reliability. UpLink introduces UpLink tokens (ULX), built on Substrate, which are used to reward users for the resources that are consumed by the network. This means that people can sell their internet to others and can use their devices as relay nodes, earning tokens in return.
+
+This proposal focuses on one of the key components of that system: the creation of a binary peer-to-peer communication protocol that mimics Substrate’s API, enabling apps to communicate with Substrate without direct access to the Internet, through a proxy. This is illustrated in Figure 1, below. A given Requester is not connected to the Internet, and rather relies on a Proxy to access the ledger. The Proxy cooperates by relaying the requests and the responses, mediating the conversation between the two. The Relay Server handles the necessary cryptographic processes to decode and interpret the contents of the request, before relaying them to the Substrate Network.
+
+![](https://user-images.githubusercontent.com/8101643/119384058-09d8aa80-bcbc-11eb-8934-7d4404a5d55c.png)
+Figure 1. Illustration of a Proxy device cooperating with a given Requester for the purposes of communicating with the Substrate Network.
+
+It’s notable that this constitutes a relaxation of the UpLink project in many ways. In its entirety, the UpLink project proposes an incentive-based multi-hop mesh network system built on Substrate, enabling devices to communicate peer-to-peer and be rewarded for their cooperation. This project, however, proposes the much simpler non-rewarded proxy alternative, which enables disconnected devices to process transactions on a Substrate network. Furthermore, it will work as a fundamental basis for the UpLink project, enabling the team to build upon this system towards its more complex counterpart. The following key differences can be highlighted:
+
+1. The network can relay content only up to one hop, meaning that there’s no actual implementation of a mesh network;
+2. It uses only Bluetooth Low Energy, instead of other connectivity technologies commonly available on the devices;
+3. There’s no incentivization model in place, meaning that the Proxy will not be rewarded for its cooperation;
+4. The Relay Server is separate from Substrate. In the final version of this project the UpLink App and Substrate Network can communicate directly, end-to-end, by integrating the necessary decoding logic directly into Substrate;
+5. The Requester is only capable of consuming a selected subset of Substrate’s functions, namely querying account balance and processing transactions.
+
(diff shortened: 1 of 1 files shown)


andre-francisco:
Hi, all,

One quick note about this PR: there was a [similar PR before](https://github.com/w3f/Open-Grants-Program/pull/412), but the reviews were so extensive that we decided to open a new one. The old PR does not really contain any necessary information, but it feels better to clarify that and avoid any further confusion. Thank you.

alxs:
Thanks for the effort you've put into this @andre-francisco. You're right, this was very well explained and easier to understand. I only have these last remarks:
- I still wonder why the Android App is the first milestone, since if I understand correctly it would depend on deliverables from later milestones? You'd usually structure the milestones such that they can be delivered in order.
- Why is the price of M2 so much higher than all the others? I think the committee might be a bit taken aback by this, especially given that its description is more scarce than that of the other milestones. I understand that it includes research into Substrate (which, also, we usually unfortunately don't fund), but even with that I have a hard time justifying the price given the three features it provides. Could you potentially reduce the price or break down/expand on your cost estimate?
- And lastly, could you provide an estimate on the duration of each milestone in terms of weeks/months, so that there is a tentative deadline for each of them?

andre-francisco:
@alxs 

Thank you for your feedback, it makes a lot of sense.

I reordered the milestones and included a table with the order of the deliveries, I think this way it should be crystal clear, both for order and duration. Also, the Communication Protocol (which is now M1) was so expensive mostly because I was playing on the safe side with respect to the security assessment; those things are usually hard to predict because you don't know what you'll find in advance, but perhaps my hesitations were exaggerated; I've reduced the total cost.

Anything else you think I should do? Your feedback is super valuable, thank you, once again!

andre-francisco:
@alxs you're right, I missed that. It's fixed now. Thank you for status update, we'll be looking forward for next steps!

alxs:
Just a note in case there's any more feedback: no need to force-push any more. This is just for the initial PR, subsequent commits are useful as they allow us to see the changes.

andre-francisco:
@alxs Understood; I'll stop rebasing in the future.

Noc2 (changes requested):
Thanks for the application. I have a few additional questions: 

- How do you plan to track the spent resources via substrate? I think this is one of the biggest problems to solve, because you quickly reach the limits of blockchains (tps) if you try to archive this in a way that can not be easily exploited. 
- In general, what do you plan to store on-chain? 
- Currently the relayer server seems to be rather centralized. Did you consider using something like an off-chain worker for this? How do you plan to incentivize people running the relayer?

andre-francisco:
@mmagician yes, it will. Citing from the proposal:

"In future reviews, this logic should be optionally integrated directly into Substrate, with the intent of having Substrate and off-the-grid devices communicate end-to-end, without the need for mediators. However, due to the complexity of that endeavour, the Relay Server will replace that setup for an initial version."

So the idea is that this communication protocol is made available as a Substrate module. That will probably be our followup for the General Grant program, since the implementation should be far more complex. For now we're using a Relay Server to replace that setup, but it's not ideal.

andre-francisco:
@Noc2 thank you for your questions, I'll leave some notes:

- How do you plan to track the spent resources via substrate? I think this is one of the biggest problems to solve, because you quickly reach the limits of blockchains (tps) if you try to archive this in a way that can not be easily exploited.

In this version the nodes are not incentivized. Citing the proposal: "It’s notable that this constitutes a relaxation of the UpLink project in many ways. In its entirety, the UpLink project proposes an incentive-based multi-hop mesh network system built on Substrate, enabling devices to communicate peer-to-peer and be rewarded for their cooperation. This project, however, proposes the much simpler non-rewarded proxy alternative [...]".

As for future versions, the incentivization model is not designed yet, but we're considering a model based on [Lot49](https://globalmeshlabs.org/files/Lot49%20Protocol%20Whitepaper%20-%20DRAFT%200.8.5_20190611.pdf), a mesh network incentivization model by goTenna. We're considering this as a followup for the General Grant program as well, or at least to have progress on this until then.

- In general, what do you plan to store on-chain?

Whatever the deployer wants. This is a generic system that enables devices to communicate, but it's up to specific implementations to decide what to do with it. That is, we work with connectivity, what that connectivity is used for is project-specific.

- Currently the relayer server seems to be rather centralized. Did you consider using something like an off-chain worker for this? How do you plan to incentivize people running the relayer?

The Relay Server will not be a part of the end system, but it will work as a means to implement the protocols and all the necessary logic until we get to the point where this is fully integrated into Substrate. It will work only for some use cases (e.g. the Relay Server can be trusted), but not for the general case; for that we'll need to go ahead with integrating the protocols as a Substrate module.

Noc2 (changes requested):
Thanks for the quick reply. Yes states channels seem to be the best solutions for this problem at the moment. You should potentially take a look at [cChannel-substrate](https://github.com/celer-network/cChannel-substrate) by celer. Just one more question from my side: Do you already have some concrete initial use cases for your technology in mind?

andre-francisco:
We'll definitely have a look at cChannel-substrate, seems interesting. Thank you for the suggestion.

Regarding the use case, yes. We're working with a major payment processing company to enable digital payment processing in markets with poor or expensive connectivity, where digital banking and the usage of digital wallets is growing exponentially. This integration with Substrate will work as a key component of such system, and our goal is also to bring these payment processing companies closer to the crypto world; this integration should achieve just that.

alxs (approved):
Thank you for the answers @andre-francisco. Looks good from my side, wish you luck with the development!

Noc2 (approved):

mmagician (approved):

andre-francisco:
@alxs 

That is great to know! Reviewing the project proposal would be tricky at the moment, since all of our resources are currently allocated. But I'll make sure to update the proposal as soon as we can.

On another note, the first deliverable is actually ready; none of the other deliverables apply right now, but would it make sense to move forward with the first deliverable? What would that entail?

alxs:
@andre-francisco sounds good!

You can certainly submit the first milestone if it's ready - the only reason I can think of why you'd want to wait would be if you expect the rest of the project to change in a way that the milestone wouldn't fit into it any longer. As long as that isn't the case, please go ahead. The process is described in the [milestone delivery repo](https://github.com/w3f/Grant-Milestone-Delivery#mailbox-milestone-delivery-process). Please take care to complete documentation and test suite for the parts of the project that you intend to deliver and then simply follow the instructions! Let us know if you have any questions.

alxs:
@andre-francisco following up here, I hope everything's going well. Do you have any news on the project, and could you give us a quick update on the timeline? Depending on where you stand, it may make sense for you to submit that [amendment](https://github.com/w3f/Grants-Program/pull/424#issuecomment-942747840) now.

alxs:
@andre-francisco is this still on your roadmap? Note that given the delay and as we haven't heard from you in a while, we'd terminate the grant if we don't hear from you in the next 2 weeks. You're welcome to apply again at any time.

PR #437, opened 2021-06-02
Add Perun channels proposal

matthiasgeihs:
Co-authored-by: Sebastian Stammler <seb@perun.network>

# Grant Application Checklist

### DIFF
--- a/applications/perun_channels.md
+++ b/applications/perun_channels.md
@@ -1,0 +1,224 @@
+# W3F Open Grant Proposal
+
+* **Project Name:** Perun Channels
+* **Team Name:** Perun@PolyCrypt
+* **Payment Address:** 0x63DcE2cB0FeA77f4E55d01569583a40e17448B4b (DAI)
+
+## Project Overview :page_facing_up:
+
+### Overview
+
+**Background.** Perun state channel technology enables fast and cheap transactions with arbitrary transaction logic on and across blockchains.
+The technology is based on peer-reviewed research that has been published at top conferences and is being implemented in the form of the open-source library _go-perun_ written in Go.
+One of its main features is its blockchain-agnostic design, which means that most parts of the library do not depend on a concrete blockchain implementation, but are rather implemented against abstract blockchain backend components. Currently the library comes with an Ethereum backend.
+
+**Proposal.** We want to provide a Polkadot backend for the _go-perun_ library in the form of a Substrate Pallet and a corresponding client integration, thereby enabling Polkadot Blockchains to benefit from the Perun channel technology.
+We view this as the first step towards enabling the full power of Perun channel technology.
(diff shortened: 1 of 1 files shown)


semuelle:
Hi Matthias, thanks for your application. Looks very good. We will look into it as soon as possible.

Noc2 (changes requested):
Thanks a lot for the application. In general it looks really good. However, I’m a little bit confused that the following is covered by the application: “Off-chain component, i.e., the Golang component that connects the on-chain logic with the client logic.”.  Isn’t this part  blockchain-agnostic anyway? And it looks to me like you are implementing the abstraction layer together with milestone 1 (Core Types). So to clarify, once this grant is finished, is it possible to use it as full payment channel solutions (without app and sub channels) or would this require additional work?

sebastianst:
Hey @Noc2 @semuelle thanks for your swift reply and feedback!

The focus of this proposal is the on-chain/conensus part in form of a Substrate Pallet _only_. We unfortunately wouldn't be able to also include the development of the go-perun _backend_ within the scope of this first proposal, as it is limited by $30k. It is correct that the off-chain _protocols_ are already implemented in an abstract way by go-perun, owing to its blockchain-agnostic architecture, but the connection between the abstraction and the Pallet still needs to be done in form of a _go-perun backend_, which we would like to build in the scope of a follow-up proposal to this one. You can have a look at our [Ethereum backend](https://github.com/hyperledger-labs/go-perun/tree/dev/backend/ethereum) to get a feel for what that means.

The payment channel protocols will already be visible in the form of tests of our Pallet. But in order to easily use payment channels via the go-perun SDK, we first need to also build the backend.

semuelle:
Have you considered applying through our [General Grants Program](https://github.com/w3f/General-Grants-Program)? It's up to US$ 100k. We usually encourage teams to apply for an Open Grant, but if you are already set on implementing both components (pallet + backend), it might give you and us more planning reliability. I would just encourage you to work on both ends simultaneously, so we're not left with a loose end in case the grant is never completed, for whatever reason.

alxs:
Sebastian has a point. While we generally encourage teams to apply to Open Grants, a General Grant might make more sense here (for the above reason, and the application also looks strong enough to me for Council approval). The only thing to consider is that the process would take a bit longer and the earliest acceptance date you'd be looking at would be in around a month. This is up to you, I'd also be willing to support the application as is and trust that you'd apply for the back end in the future.

Noc2 (approved):
Thanks for the response. I think it makes not much difference for us, if we support it via open or general grants. In both cases, we rely on you to actually finish the project. I’m therefore going to accept this grant.

semuelle (approved):
In that case, I won't dissent. :)

mmagician (approved):
Thanks for putting in the extra effort to polish the deliverables, the level of detail is much deeper than the initial proposal you shared by email! I'm happy to proceed.
Also you can safely ignore my last email :)

sebastianst:
Awesome, thanks a lot! We're looking forward to working on the project 🎉

PR #453, opened 2021-06-11
Decentralised reputation systems and social networks

DarkingLee:
# Grant Application Checklist

### DIFF
--- a/applications/ZeroDAO_Network.md
+++ b/applications/ZeroDAO_Network.md
@@ -1,0 +1,243 @@
+# W3F Open Grant Proposal
+
+- **Project Name:** ZeroDAO Network
+- **Team Name:** ZeroDAO
+- **Payment Address:** 0xEBCDa7c73EB5Dd7a4314cFf395bE07EB1E24c046
+
+## Project Overview 📄
+
+### Overview
+
+We define ZeroDAO as a public resource, including a social network, a reputation system. the ZeroDAO social network solves the incentive dilemma that currently exists in blockchain social networks, while incentivizing good behavior makes good behavior disappear. Imagine what Twitter would look like if you could get $1 for posting a tweet.  [Two-factor theory](https://en.wikipedia.org/wiki/Two-factor_theory) even concludes that security, salary, fringe benefits, good pay is not **[Motivators](https://en.wikipedia.org/wiki/Motivation)** but **Hygiene factors**. Hygiene factors that do not give positive satisfaction or lead to higher motivation.
+
+ZeroDAO social network solves the incentive dilemma by amplifying social motivation and internalizing external motivation.
+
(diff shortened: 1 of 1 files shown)


Noc2 (changes requested):
Thanks for the application. This looks really interesting and something we would like to support. I actually wanted to write an RFP for an on-chain reputation system, which for example could be used to judge the reputation of parachains/dapps etc. and therefore help others to make the right investment decisions. I have a few questions: 

- How do you ensure that the hall of fame isn’t malicious. Did you choose a random number of “famous” people? Ideally this group is also somehow decentralized to increase the risk of collusion. 
- You say TIR is difficult to compute, how do you plan to incentivise people to actually compute it in a decentralized way.   
- For the other implementation, it seems you used thegraph. Potentially the following is helpful for you: https://github.com/Joystream/hydra

DarkingLee:
@Noc2  Thank you for your reply.

1. In socircles, I combined several centrality values to select "famous" people. We will implement community elections in the next milestone, which will be decentralised and secure. In socircles, only 20 users out of 100,000 were selected and the system already works well. That means that on average 5,000 users oversee a famous people, which is safe. But elections need to be secure in combination with reputation to ensure that an attacker without a reputation, cannot attack the election. However at the moment we do not have a reputation. Also to take into account options such as MACI or privacy computing, which I note some teams in the ecology are already working on. Given the above points, I think it would be better to implement it in the next milestone.

2. TIR is really difficult to compute on-chain. But it is much easier to compute off-chain. Users updating the results to the chain will get a percentage of the user's social currency as a fee.

3. Thanks for sharing.

Noc2 (approved):
Thanks for the quick reply. I mostly ask about TIR, because it seems it’s currently not part of your deliverables. But I guess you want to look into it in the future as part of “Optimize economic system design”. In general I’m happy to support the proposal and I will share the application with the rest of the team.

BenWhiteJam (approved):

DarkingLee:
Hi @Noc2  thanks for your great effort! As you said, we have plans to provide a client for TIR. But we are not ready yet. For now we are only verifying it on-chain.

alxs (approved):

alxs:
Hey @DarkingLee, care to share a quick status update along with an ETA if possible? We expected to receive your first delivery a bit over 2 months ago now. As long as you're actively working on the project, delays aren't a big problem, but please keep us in the loop.

Note that if we don't hear from you within the next two weeks, we'll assume you're no longer interested and terminate the grant.

DarkingLee:
Hey @alxs, Very sorry for the delay in delivery. Due to the completeness of the project we have added a seed selection algorithm which has taken a long time. Development is now almost complete and we are preparing to refine the documentation for delivery in the next 2 weeks at the latest.

alxs:
@DarkingLee ping.

DarkingLee:
@alxs Delivery this week.

PR #454, opened 2021-06-14
zkSNARK Plonk

NoCtrlZ:
# Grant Application Checklist

### DIFF
--- a/applications/zk-plonk.md
+++ b/applications/zk-plonk.md
@@ -81,23 +81,26 @@ We plan to provide a `plonk` pallet that allows Substrate-based blockchain to ex
 
 ### Overview
 
-* **Total Estimated Duration:** 4 months
+* **Total Estimated Duration:** 3 months
 * **Full-Time Equivalent (FTE):**  1 FTE
-* **Total Costs:** 40000 DAI
+* **Total Costs:** 30000 DAI
 
 ### Milestone 1 Example — Implement Substrate Modules
 
-* **Estimated Duration:** 4 months
+* **Estimated Duration:** 3 months
 * **FTE:**  1
-* **Costs:** 40000 DAI
+* **Costs:** 30000 DAI
 
 | Number | Deliverable | Specification |
 | -----: | ----------- | ------------- |
 | 0a. | License | Apache 2.0 / MIT / Unlicense |
 | 0b. | Documentation | We will provide both inline documentation of the code and a basic tutorial that explains how a developer builds a circuit and a user prove their computation through the circuit. |
-| 0c. | Testing Guide | Core functions will be fully covered by unit tests to ensure functionality and robustness. In the guide, we will describe how to run these tests. |
+| 0c. | Testing Guide | Core functions will be fully covered by unit tests and audit to ensure functionality and robustness. In the guide, we will describe how to run these tests. |
 | 0d. | Article/Tutorial | We will publish an article/tutorial/workshop that explains
-| 1. | plonk zkSNARK pallet | We will create a set of plonk-based zkSNARK libraries that allow a developer to build their own circuit and a user to prove their computation validity. |  
+| 1. | make plonk compatible | The dusk-network plonk is compatible with `no-std` so we are going to modify attributes according to [parity-codec](https://github.com/paritytech/parity-scale-codec) and `Rng` to be compatible with　Substrate environment. |
+| 2. | implement zkSNARK plonk pallet | We will create a set of plonk-based zkSNARK libraries that allow a developer to build their own circuit and a user to prove their computation validity. |  
+
+This zkSNARK plonk is based on [dusk-network plonk](https://github.com/dusk-network/plonk) library.
 
 ## Future Plans
 
--- a/applications/zk-plonk.md
+++ b/applications/zk-plonk.md
@@ -1,0 +1,112 @@
+# W3F Open Grant Proposal
+
+* **Project Name:** zk plonk
+* **Team Name:** Plasm Network (Shinsaku Ashizawa, Sota Watanabe)
+* **Payment Address:** 0xb82EdE43D03fD23dcdb2d066720b3E77F3bedf6d
+
+## Project Overview :page_facing_up:
+
+We have been working on Zk projects and scalability solutions and now we would like to implement Zk plonk pallet.
+
+The plonk is called `universal zkSNARK` because of two reasons. The verification times are constant and original reference string can be used to build proofs with any type of circuit. These features bring significant benefits to both users and developers. For example, the verification time is the same so users don't have to wait so long even when they prove complicated proof, and original reference string can reuse so developers don't have to do trusted setup for each circuit. The plonk will be a core technology in terms of scaling and privacy so we should support.
+
+### Overview
+
+Through this grant, we are going to make a plonk pallet in order for the developer to implement plonk on substrate easily. We are working on scalability solutions but currently, the substrate doesn't support zkSNARK pallet so we think it's a issue to fix. As aforementioned, zkSNARK will be a core technology in blockchain area and especially plonk is cutting edge technology so we are excited to implement this as pallet.
+
+### Project Details
+
+We'd like to implement the plonk library as a pallet in order for developers to customize circuits and use the plonk protocol.
+
+The following diagram is the libraries we are going to implement.
+
+* Polynomial commitment
+* Circuit builder
+* Generate proof and verify keys
+* Verify proof
+
+### Ecosystem Fit
+
+According to Web3 Foundation, there are at least 2 different teams that work on ZK technologies.
+- [Zeropool](https://github.com/zeropoolnetwork)
+- [Glacier](https://github.com/gbctech)
+
+Glacier is building a Distaff VM for zk-STARK proof generation and verification that are used to make  private smart contracts and private credential verifications. The difference between us is that we are making a zkSNARK pallet and they are making a VM which supports STARKs. In terms of Zeropool, they are making private transactions contract pallet using bellman groth16 protocal and we are making zkSNARK libray using plonk.
+
+We believe zkSNARK with plonk will be core technology of next blockchain area. That allows Substrate to protect the privacy and scale on the huge number of transactions without decreasing the security level.
+
+## Team :busts_in_silhouette:
+
+### Team members
+
+* Shinsaku Ashizawa
+* Alexsandr Krrupenkin
+* Sota Watanabe
+
+### Contact
+
+* **Contact Name:** Shinsaku Ashizawa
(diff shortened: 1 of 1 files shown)


Noc2 (changes requested):
Thanks for the application. I think in general we are highly interested in this. Are you going to use the [Pure Rust implementation by dusk-network](https://github.com/dusk-network/plonk)? What exactly do you need to do as part of deliverables in order to implement this into substrate? 4 months seems to be a lot for this. Are you also planning to audit the implementation?

NoCtrlZ:
@Noc2 
Thank you for the questions 😎

> I think in general we are highly interested in this.

Thank you.
I am really honored.
> Are you going to use the Pure Rust implementation by dusk-network?

Exactly.
> What exactly do you need to do as part of deliverables in order to implement this into substrate?

The dusk-network plonk is compatible with `no-std` so we are going to change some attributes according to [parity-codec](https://github.com/paritytech/parity-scale-codec) to be compatible and modify `Rng` to be compatible with wasm.
After that step, we are going to implement plonk as a pallet.
> 4 months seems to be a lot for this.

Okay.
I will buy a new keyboard to type fast and it reduces a month so it takes 3 months totally.
> Are you also planning to audit the implementation?

Absolutely!

Please feel free to ask any questions.
Thank you!

Noc2:
Thanks for the quick reply. Could you update the application accordingly (concrete deliverables, potentially the audit, etc)?

NoCtrlZ:
Hi @Noc2 
I updated it according to your review 😎
Please confirm.
Thank you!

NoCtrlZ:
Hi @mmagician
Thank you for the questions.

> modify attributes according to parity-codec and Rng to be compatible with Substrate environment.
> I would like to understand a little bit better what you mean by the above.

In order for the pallet to work on resource-constrained execution environments like Substrate runtime, attributes should be modified in accordance with the SCALE codec.
The parity-codec is a high-performance encoding and decoding data library.
In terms of `Rng`, there are some versions of `Rng` that can’t be compiled to wasm so we need to research and make it stable if it needs.
> Did I understand correctly that you want to provide the ability to both creating the proofs as well as verifying them from your pallet?

Creating the proofs is done off-chain because it needs secret input and it shouldn’t be public, and that process consumes a bunch of computation.
Proofs are created off-chain and verify them on-chain through our verifying pallet.
> Lastly, could you provide a list of extrinsics that would be callable via your pallet? This would later allow us to better evaluate your delivery against the original contract.

This pallet allows us to do building circuits, creating proofs, and verifying proofs.
> I would like to kindly ask you to include the requested information in your application.

I updated accordingly.
Please confirm.
Thank you!

applications/zk-plonk.md
```diff
@@ -97,10 +97,14 @@ We plan to provide a `plonk` pallet that allows Substrate-based blockchain to ex
...
-| 2. | implement zkSNARK plonk pallet | We will create a set of plonk-based zkSNARK libraries that allow a developer to build their own circuit and a user to prove their computation validity. |  
+| 1. | make plonk compatible | The dusk-network plonk is compatible with `no-std` so we are going to modify attributes according to [parity-codec](https://github.com/paritytech/parity-scale-codec) and `Rng` to be compatible with　Substrate environment. This step allows this pallet to work on resource-constrained execution environments like Substrate runtime, attributes should be modified in accordance with SCALE codec and some versions of Rng can’t be compiled to wasm so we need to research and make it stable as necessary. |
+| 2. | implement zkSNARK plonk pallet | We will create a set of plonk-based zkSNARK libraries that allow a developer to build their own circuit and a user to prove their computation validity. Verifying proofs are done by on-chain. Creating the proofs are done by off-chain. |
```
mmagician, line 101: off chain - meaning off-chain worker? or a separate tool?

applications/zk-plonk.md
```diff
@@ -0,0 +1,119 @@
...
+| 0d. | Article/Tutorial | We will publish an article/tutorial/workshop that explains
+| 1. | make plonk compatible | The dusk-network plonk is compatible with `no-std` so we are going to modify attributes according to [parity-codec](https://github.com/paritytech/parity-scale-codec) and `Rng` to be compatible with　Substrate environment. This step allows this pallet to work on resource-constrained execution environments like Substrate runtime, attributes should be modified in accordance with SCALE codec and some versions of Rng can’t be compiled to wasm so we need to research and make it stable as necessary. |
+| 2. | implement zkSNARK plonk pallet | We will create a set of plonk-based zkSNARK libraries that allow a developer to build their own circuit and a user to prove their computation validity. Verifying proofs are done by on-chain. Creating the proofs are done by off-chain. |  
+
+This zkSNARK plonk is based on [dusk-network plonk](https://github.com/dusk-network/plonk) library.
+This zkSNARK plonk pallet provides us following function.
+- Building circuits
```
mmagician, line 105: You're planning to build circuits through your pallet? This sounds like you're planning to make it interactive, which doesn't make much sense to me, since circuits needed to be **programmed** by the user. As far as I'm aware, this involves writing code with the help of some tools (like [circom](https://docs.circom.io/), or dusk-plonk would be the fit here I guess) that is later compiled. Please expand on how your pallet would support building circuits

applications/zk-plonk.md
NoCtrlZ: Creating the proofs is done by an off-chain worker.

applications/zk-plonk.md
NoCtrlZ: Hi @mmagician 
Thank you for asking me for details.

> You're planning to build circuits through your pallet? This sounds like you're planning to make it interactive, which doesn't make much sense to me, since circuits needed to be programmed by the user.

Yes.
The user(Substrate developer) can build their own programmable circuit.
Let's say they want to build a circuit of privacy transfer, they would build a circuit on-chain(Substrate node) that checks the validity of transactions, and the transactor sends transactions through the off-chain worker.
It's a similar structure with circom which deploys verifier contract and this is the `verify pallet` of this project.
The circom creates proof using the off-chain library and this is the `create proof` off-chain worker of this project.

mmagician: If you don't mind I'll lean on the example from circom again. From [this tutorial](https://blog.iden3.io/first-zk-proof.html) section 2.1, a user needs to **write code** in order to build their circuit.

So what you're saying is that if a developer wants to build their own circuit, they need to have a separate pallet with the logic for their own circuit? Because, as far as I understand, each circuit will be customised for the developer's own purpose and it's difficult to make it generic.

I could imagine that if there is a common interface, an off-chain worker could be called to build the circuit from whatever code the user has written. Could it be generic enough?

Well if that's the case and I understood your last comment, then the building of the circuit will not happen in your pallet, but rather each developer has to do it independently. And then they can call your pallet's `create proof` method (assuming there is some standard interface in place) to take whatever circuit they built, supply the secret inputs, and compute the proof. Is my understanding correct?

NoCtrlZ: Thank you for the questions.
> If you don't mind I'll lean on the example from circom again. From this tutorial section 2.1, a user needs to write code in order to build their circuit.

This plonk pallet provides a circuit template library.
A user writes code on the provided circuit template library so it doesn't have to have a separate library by itself in the same way as a circom user doesn't need to implement any kind of library just defining constraint.
> they need to have a separate pallet with the logic for their own circuit?

No, because this plonk pallet provides a circuit template library.
> each circuit will be customised for the developer's own purpose and it's difficult to make it generic.

Each circuit will be customized using a circuit template library and a user define the constraint themselves.
> I could imagine that if there is a common interface, an off-chain worker could be called to build the circuit from whatever code the user has written.

Sorry I couldn't understand this.
The circuit can't be added after a developer deploys the Substrate node.
> the building of the circuit will not happen in your pallet

The building of the circuit will happen in this plonk pallet using a circuit template library.
A user defines the circuit as following through the plonk pallet.

```rust
impl Circuit for TestCircuit {
    const CIRCUIT_ID: [u8; 32] = [0xff; 32];
    fn gadget(
        &mut self,
        composer: &mut StandardComposer,
    ) -> Result<(), Error> {
        let a = composer.add_input(self.a);
        let b = composer.add_input(self.b);
        // Make first constraint a + b = c
        composer.poly_gate(
            a,
            b,
            composer.zero_var(),
            BlsScalar::zero(),
            BlsScalar::one(),
            BlsScalar::one(),
            BlsScalar::zero(),
            BlsScalar::zero(),
            Some(-self.c),
        );
        // Check that a and b are in range
        composer.range_gate(a, 1 << 6);
        composer.range_gate(b, 1 << 5);
        // Make second constraint a * b = d
        composer.poly_gate(
            a,
            b,
            composer.zero_var(),
            BlsScalar::one(),
            BlsScalar::zero(),
            BlsScalar::zero(),
            BlsScalar::one(),
            BlsScalar::zero(),
            Some(-self.d),
        );

        let e = composer.add_input(self.e.into());
        let scalar_mul_result = composer
            .fixed_base_scalar_mul(e, dusk_jubjub::GENERATOR_EXTENDED);
        // Apply the constrain
        composer.assert_equal_public_point(scalar_mul_result, self.f);
        Ok(())
    }
    fn padded_circuit_size(&self) -> usize {
        1 << 11
    }
}
```
A user circuit is `TestCircuit` and the `Circuit` struct is from this pallet.
A user can build a circuit by extending provided `Circuit` struct and this is a generic circuit template library.
And from the next line, a user defines the constraint.
I hope this example helps you to understand well.
Sorry that was such a long comment 🙇‍♂️

mmagician (approved):
Okay, that indeed clarifies the idea for me. Thanks a lot for the explanation!

alxs (approved):

NoCtrlZ:
Hi @Noc2 
> Also let me know if we should help you with the audit and potentially try to find a company, etc.

I'd like to know what kind of help you provide about the audit.

Noc2 (approved):

Noc2:
> Hi @Noc2
> 
> > Also let me know if we should help you with the audit and potentially try to find a company, etc.
> 
> I'd like to know what kind of help you provide about the audit.

We could potentially help you to find a company. It’s currently pretty difficult to find good audit companies.

alxs:
@NoCtrlZ any updates, are you still interested in working on this?

NoCtrlZ:
Hi @alxs 
It's almost completed.
https://github.com/PlasmNetwork/plonk
The prover, circuit builder and setup, verify functions work correctly!
https://github.com/PlasmNetwork/plonk/blob/master/src/tests.rs#L207
I am working on tutorial and documentation so I will deliver with in 2 weeks.
Sorry for late.

NoCtrlZ:
Hi there
I created the delivery PR following!
Please confirm!
https://github.com/w3f/Grant-Milestone-Delivery/pull/326
Thank you!

PR #455, opened 2021-06-15
Create NuLink.md

pawnz0:
# Grant Application Checklist

### DIFF
--- a/applications/NuLink.md
+++ b/applications/NuLink.md
@@ -18,7 +18,9 @@ The main components of the project are:
 ### Project Description
 
 + In the NuCypher Network, all the cryptography operation are currently performed in an off-chain side channel. We can reuse this side channel in our project. We only need to take care of the on-chain operation from Ethereum to Polkadot. The on-chain operation mainly include two parts: Stakers need to register and get rewards on-chain, while Alice need to grant access to Bob and pay fees to stakers on-chain. NuLink would focus on migrating the on-chain operation from Ethereum to Polkadot parachain.
-+ We will develop a mirror pallet which can retrieve the information of stakers and bonding workers from Nucypher contracts in Ethereum. At current stage we consider constructing a watcher network maintained by NuCypher team. The watcher network would relay such information to the mirror pallet.  In the future we will use ETH-Polkadot bridge such as Snowbridge to retrieve such information from Ethereum.
++ We will develop a mirror pallet which can retrieve the information of stakers and bonding workers from Nucypher contracts in Ethereum. At current stage we consider constructing a watcher network managed by NuCypher team. The watcher network would relay such information to the mirror pallet. 
++ The watcher network would be constructed by a certain amount(say N) of watcher nodes.  And their corresponding public key would be recorded in the mirror pallet when first deployed.  The update request must obtain more the 2/3 signature of the total watcher nodes to be valid. The mirror pallet would check the signature  when receiving the update request.  The locking period of Ursulas network is 7 days, so the watcher network would sending the update request in the same pace.
++  In the future we may use ETH-Polkadot bridge such as Snowbridge to retrieve such information from Ethereum instead of the watcher network. The details would be updated once we finish the design.
 + We would also develop a policy management pallet which can hold policy fees and distribute them in Polkadot parachain. Through this pallet, Alice in Polkadot could issue a policy in polkadot parachain and pay the policy fee in Polkadot parachain. And the worker of Ursulas could withdraw the reward in Polkadot parachain.  
 
 
@@ -57,8 +59,10 @@ https://github.com/pawnz0/NuLink
 | 0a. | License | Apache 2.0 |
 | 0b. | Documentation | We will provide both documentation of the code and a basic tutorial that explains how a user can use the pallet. |
 | 0c. | Testing Guide | This milestone will have unit-test for each function to ensure functionality. In the guide we will describe how to run these tests. |
-| 1.     | UpdateStakers   | This function would provide the functionality of updating the information of current stakers and bonding workers of Ursulas network. |
-| 2.     | GetActiveStakers | This function would return a list of active stakers by random sampling. |
+| 1.     | SetWatchers   | This function would record the public key of the watcher nodes and would be executed when the mirror pallet first deployed. |
+| 2.     | ValidateWatchers | This function would check the signature in updating request. It would return true only when more than 2/3 watcher nodes sign the updating request. |
+| 3.     | UpdateStakers   | This function would provide the functionality of updating the information of current stakers and bonding workers of Ursulas network. |
+| 4.     | GetActiveStakers | This function would return a list of active stakers by random sampling. |
 
 ### Milestone 2 Implement Policy Management Pallet
 * **Estimated Duration:** 1 month
@@ -79,4 +83,4 @@ https://github.com/pawnz0/NuLink
 
 ## Additional Information :heavy_plus_sign:
 
-We have discussed this project  with NuCypher core team and get some technical support from NuCypher team.
+We have discussed this project  with NuCypher core team and get some technical support from NuCypher team. The client of watcher nodes would be implemented indepently and not funded by this proposal. We would develop the watcher network along with NuCypher team to provide such functionality. 
--- a/applications/NuLink.md
+++ b/applications/NuLink.md
@@ -1,0 +1,82 @@
+# Open Grant Proposal
+* **Project:** NuLink
+* **Proposer:**  Pawn
+* **Payment Address:**  0xf7410438ead9e89EEd5ca6e61a11E862C297ca0e
+
+## Project Overview :page_facing_up:
+### Overview
+
+Our project NuLink is trying to bridge the NuCypher Network to Polkadot Ecosystem. The [NuCypher Network](https://docs.nucypher.com/en/latest/index.html) is a decentralized network of nodes that perform threshold cryptography operations serving users with secrets management and dynamic access control; in particular, NuCypher currently offers a threshold proxy re-encryption solution that enables secure data sharing which can be used to construct content management platforms, secure messaging, encrypted NFTs, and many other applications.
+
+Through our project, users and applications from Polkadot could take advantage of NuCypher Network. For example, a user Alice in Polkadot could share her private data to another user Bob using the integration from our project, without worrying information leaking to any third party, thanks to NuCypher cryptographic assurance.
+
+The main components of the project are:
+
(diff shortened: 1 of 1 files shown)


Noc2 (changes requested):
Thanks for the application. I’m just putting a reference here to your previous application: https://github.com/w3f/Open-Grants-Program/pull/345 Did you talk to nucypther directly? How exactly would a watcher network work? This sounds highly centralized to me. To be honest, I personally think that your previous application was more useful for us, since this would basically implement the same but with an additional (potentially centralized) component integrated.

pawnz0:
Thanks, Noc. Yes, we have met with Nucypher core team several times and they suggest us to reuse their proxy network Ursulas(since there is already a mature workers network, we don't need to reconstruct a similar one in Polkadot). And the watcher network is the current choice. Later we would shift this to Snowbridge or other ETH-Polkadot bridge to make it decentralized. In our last proposal,  your teammate suggest us to bridge Nucypher network to Polkadot instead of rebuilding a whole Nucypher network in Polkadot. We re-consider his suggestion and discussed with Nucypher team to make this new proposal. If you have any other issue please let me know.

Noc2 (changes requested):
Thanks for the quick reply. I personally disagree with Jeff’s suggestion regarding bridging. I think a direct integration has a lot of advantages and from my experience the grants committee is currently also reluctant to fund bridges, especially if they contain rather centralized components. Independent of it I will share the application with the rest of the team, but I would actually be much more interested in your previous application.

cygnusv:
Hey @Noc2, this is David from the NuCypher team. @pawnz0's team talked to us about this proposal and we gave them some feedback and recommendations. Their original idea implied replicating the NuCypher Network from scratch in Polkadot, which would be by itself a daunting task. But, putting aside this effort, it would also imply attracting, on-boarding and providing support to a collective of node operators to stake in this mirror network, essentially, replicating the supply side of the NuCypher network. After doing this with over 2,000 nodes, we learned how challenging and time-consuming this task is as well. 

Instead, if this project assumes that the supply side is covered, by bridging it from the existing NuCypher network, it can focus exclusively in what users and applications on the Polkadot side need. I do agree that there's a centralized component in this proposal, but in my view it can serve as a first step while ETH-Polkadot bridges consolidate.

alxs:
Hey @cygnusv, thanks for checking in and adding your perspective. Even though I can see where @Noc2 is coming from, what you're saying also makes sense to me and I agree this can indeed serve as a first step.

However I'm also wondering how the watcher network would work exactly, and I'm concerned it's currently not included in the deliverables as, the way I see it, the rest of them are of no use without it.

Nevertheless I also wonder whether this is needed at all. I.e. the watcher network would basically be a number of nodes keeping tabs on the NuCypher smart contracts and then submitting changes, right? So you could potentially simply use something like [ChainBridge](https://moonbeam.network/community/projects/chainbridge-by-chainsafe/) which should soon be available on Kusama anyway. Or how would this watcher network be different from a PoA bridge?

I'm betting by the time this is finished there's already an Ethereum bridge deployed somewhere or close to being deployed, so maybe you can aim straight for this option. Though in all fairness I'm probably being too optimisic.

And lastly: would it be possible to supply both blockchains using the same network of Ursulas? Couldn't this work cross-chain?

cygnusv:
> And lastly: would it be possible to supply both blockchains using the same network of Ursulas? Couldn't this work cross-chain?

Yes! That's why, IMHO, an approach that tries to make use of NuCypher's existing supply of Ursulas shows more potential. NuCypher's ERC20 token and staking information currently lives on Ethereum, but that's something that's part of the work token model, and not important for users and applications. In fact, from that perspective, our network of Ursulas can be blockchain-agnostic, assuming projects like NuLink enabling native collection of service fees in other L1s. In Ethereum, we collect fees on ETH, but here it could be in DOT or even in a stablecoin (I guess depending on which parachain NuLink ends up targeting).

pawnz0:
David, Thanks for your clarification. 

NuLink is exactly trying to supply Polkadot with Ursulas network. And for the watcher network, we original design it as a POA bridge. Some selected relayer nodes would have the authority to update the newest state under multisignature mechanism. We are not aware that Chainbridge would be available on Kusama soon. We could directly use Chainbridge to provide information for mirror pallet instead of the watcher network if you think this is a better option.

alxs:
@pawnz0 @cygnusv how about we discuss this a bit more extensively in a quick call? I'm OOO for the next ~2 weeks, but both tomorrow afternoon or Friday would still work well for me (CET). Feel free to send me an invite to `aleixo@web3.foundation`.

mmagician:
> We will develop a mirror pallet which can retrieve the information of stakers and bonding workers from Nucypher contracts in Ethereum. At current stage we consider constructing a watcher network maintained by NuCypher team. The watcher network would relay such information to the mirror pallet

In general I'd be fine with a **centralised** watched network, as long as it's **trustless**. Basically, what I'm looking for would be a way to ensure correctness and non-forgeability of the data supplied to substrate. Could the pallet-relayer connection support such functionality? Ideally, anyone would be able to just spin up such a watcher and have it supply data to the mirror module, e.g. in case the watcher maintained by a third party is unavailable.

pawnz0:
We have update the proposal and give the design of the watcher network.

mmagician:
So just to re-iterate on my last questions: in case a new user wants to run their own watcher, how easily could they get their PK included in the mirror pallet's storage?

pawnz0:
> So just to re-iterate on my last questions: in case a new user wants to run their own watcher, how easily could they get their PK included in the mirror pallet's storage?

They can not. The watcher network is designed as a stable set. The watcher list is set up when the mirror pallet is deployed. In the future we would substitute the watcher network to a ETH bridge. The mirror pallet would be able to interact with the bridge to get update information of Ursulas nodes then.

mmagician:
But then, isn't verifying 2/3 of the signatures pointless, if these are pre-selected (i.e. centralised?) watchers?

pawnz0:
> But then, isn't verifying 2/3 of the signatures pointless, if these are pre-selected (i.e. centralised?) watchers?

Yes, they are pre-selected nodes. NU cypher team would manage this network. The signature mechanism is in case some nodes is offline.

mmagician:
Just to make sure I understand this correctly: Alice would be able to decide if she wants to set up the policy on Ethereum or Polkadot, and con

cygnusv:
Yes, the idea is that Alices can decide to pay for policies either on the Ethereum side or in the Polkadot side; it's essentially opening a new market for NuCypher workers that currently only can get fees from Ethereum. 

I don't think it has to be necessarily opt-in, as the work itself (i.e. the distributed re-encryption process) is always off-chain and blockchain-agnostic. It may be simpler to just show the workers that Polkadot is an additional source of revenue for them, than introducing an opt-in process that may produce more UX friction for workers .

mmagician (approved):
Ok, that sounds good. I think I understand your proposal a bit better now and will be happy to support the project!

mmagician (changes requested):
Btw, I still don't think the following makes sense:
> The signature mechanism is in case some nodes is offline.

It's good to have more than one node for availability. But if they are controlled by the same entity, it doesn't matter if there is a 2/3 majority, of a leader-follower model (Raft?) or even just a round robin model that allows one watcher to interact with the network.

alxs (approved):
Happy to support the current approach. Keep in mind that:
- Seeing as you want to go for pallets as opposed to smart contracts: in case you don't find a parachain project willing to integrate your pallets, it is also an option to launch your own.
- As previously discussed, whether you need to implement and deploy the watcher network at all depends on whether an Ethereum bridge is available on some parachain soon enough. I wouldn't blindly wait for this to happen, but you probably want to look into it.

cygnusv:
> It's good to have more than one node for availability. But if they are controlled by the same entity, it doesn't matter if there is a 2/3 majority, of a leader-follower model (Raft?) or even just a round robin model that allows one watcher to interact with the network.

We at NuCypher would also prefer a more decentralized approach for the watcher nodes. I think current proposed design allows for it anyway, so if NuLink can clarify that part, it would be great.

Noc2 (approved):

pawnz0:
> It's good to have more than one node for availability. But if they are controlled by the same entity, it doesn't matter if there is a 2/3 majority, of a leader-follower model (Raft?) or even just a round robin model that allows one watcher to interact with the network.

Thanks for the comments. I think maybe we can invole more orgnazations (NuCypher, Web3 foundation, etc.) to maintain this network together to make the watcher network more decentralized. All the watcher nodes can sumbit update request to the mirror pallet seperately with signature. And when the mirror pallet accumulate enough request(reaching the 2/3 threshold), the new state can then be recorded in the parachain.

pawnz0:
> Happy to support the current approach. Keep in mind that:
> 
> * Seeing as you want to go for pallets as opposed to smart contracts: in case you don't find a parachain project willing to integrate your pallets, it is also an option to launch your own.
> * As previously discussed, whether you need to implement and deploy the watcher network at all depends on whether an Ethereum bridge is available on some parachain soon enough. I wouldn't blindly wait for this to happen, but you probably want to look into it.

Thanks a lot for your suggestions. We will keep tracking the newest update of the bridge. Also we would try to launch a parachain by ourselves if we can not find one.

mmagician (approved):
Ok, sounds good. You can also update the deliverables during the grant should anything change.

alxs:
@pawnz0 would you mind sharing a quick status update, along with an ETA if possible? The [project repo](https://github.com/pawnz0/NuLink) looks inactive and we were hoping to receive your first delivery a bit over 2 months ago now.

As long as you're actively working on the project, delays aren't a big problem, but please keep us in the loop. If you expect the delivery to take longer than 2-3 more weeks, you'll need to submit an amendment.

Note that if we don't hear from you soon, we will assume you're no longer interested and terminate the agreement.

alxs:
@pawnz0 it's good to see that you're still working on this project. When do you expect to submit the first milestone? Note that if you've deviated from the original deliverables or funding requirements, you can always submit an [amendment](https://github.com/w3f/Grants-Program#changes-to-a-grant-after-approval).

alxs:
@pawnz0 can you please give us some sort of update? Given the current delay of over 6 months, you should submit an amendment to update the timeline.

pawnz0:
Hello, sorry for the delay. And we are ready for delivery of both milestones now. We will submit the delivery soon next week. You can check our project [here](https://github.com/NuLink-network/nulink-chain).

alxs:
That's great news. In that case, you can just deliver the milestones without an amendment.

PR #463, opened 2021-06-21
Yatima W3F Open Grant application

johnchandlerburnham:
# Grant Application Checklist

### DIFF
--- a/applications/yatima.md
+++ b/applications/yatima.md
@@ -1,0 +1,443 @@
+# W3F Open Grant Proposal
+
+* **Project Name:** Yatima
+* **Team Name:** [Yatima Inc](https://github.com/yatima-inc/yatima).
+* **Payment Address:** 3N2bLsFX6UtJ3MXdnVvm87dHaQTPutDiqA
+
+## Project Overview :page_facing_up:
+
+### Overview
+
+* Summary: A lambda-VM and programming language for Substrate chains, with
+  on-chain formal verification and IPFS package management.
+
+* Description: Yatima is a dependently typed functional programming language
+  implemented in Rust featuring substructural types, metaprogramming, and
+  content addressing via IPFS.
+
+* Relevance to Substrate: The language core, including parsing, elaboration,
+  serialization/deserialization, typechecking, evaluation, etc. can compile to
+  Substrate compatible WASM (`no_std` via Substrate's [`sp_std` standard library](https://docs.rs/sp-std/3.0.0/sp_std/index.html)), and thus can be
+  utilized within a Substrate Pallet.
+
+* Our interest: In order for decentralized technologies to gain widespread
+  adoption, users have to trust that the infrastructure they are using is safe,
+  and that they are not just replacing counterparty risk with technical risk. We
+  believe that functional programming and formal proofs offer a possible path
+  towards increasing the safety and reliability of all software, not just in
+  blockchain.
+
+### Project Details
+
+Many blockchain projects are exploring strongly-typed functional programming as
+a paradigm for writing smart contracts and DApps. Examples include [Cardano's
+Plutus](https://developers.cardano.org/en/programming-languages/plutus/overview/), [Tezos' Michelson](https://wiki.tezosagora.org/learn/smartcontracts/michelson), [Kadena's Pact](https://github.com/kadena-io/pact) and [Digital Asset's DAML](https://daml.com/). The type safety and
+composability of functional programs make them an exciting option for
+writing high-assurance smart contracts.
+
+Yatima aims to help develop the functional programming ecosystem on Substrate
+and the Polkadot ecosystem. We have developed a graph-based runtime for the Yatima language
+with fully lazy evaluation, sharing, and whole program optimization. This runtime could also 
+be targeted by other functional languages. The Yatima runtime is
+similar in many respects to Haskell Core, or the [GRIN
+IR](https://grin-compiler.github.io/) but is implemented in deterministic
+`no_std` Rust, which allows it to run virtually anywhere, including the web,
+WASM blockchains, or even embedded systems (in theory).
+
+Yatima also implements a powerful dependent and substructural type-system,
+similar to Idris 2. This enables formal verification and theorem proving
+(although crucial features for practical proving such as inference and tactics
+are not yet implemented). Proofs allow for many useful optimizations in the area
+of smart contracts, such as static detection of arithmetic overflow or division
+by zero errors, or tracking of resources and permissions via an effect system.
+Furthermore, since Yatima can be implemented in itself, there is a possiblity of
+having on-chain trustless proof verification.
+
+Lastly, Yatima implements content-addressing, allowing all programs and packages
+to be serialized into IPLD objects and shared over the peer-to-peer IPFS network
+(or via libp2p).
+
+#### Deliverables
+
+- A `sp_ipld` crate for interacting with IPFS CID, Multihash and IPLD based on
+  [`ipfs-rust/libipld`](https://github.com/ipfs-rust/libipld), but which is
+  Substrate pallet compatible. This is a necessary dependency for Yatima Core,
+  but is an independent deliverable that could be useful for many other
+  Substrate projects.
+
+- A `sp_im` crate for creating immutable datastructures, based on
+  [`bodil/im-rs`](https://github.com/bodil/im-rs), but which is
+  Substrate pallet compatible. This is a necessary dependency for Yatima Core,
+  but is an independent deliverable that could be useful for many other
+  Substrate projects.
+
+- A `sp_yatima` crate integrated with `sp_ipld`, `sp_im` and `sp_std` that
+  packages the `yatima_core`
+  [crate](https://github.com/yatima-inc/yatima/tree/main/core) and a
+  `yatima-node` repository based on the
+  [`substrate-node-template`](https://github.com/substrate-developer-hub/substrate-node-template)
+  showcasing how `sp_yatima` can be imported into a Substrate Node as a pallet.
+
+This proposal does *not* include in its scope:
+  - Implementing a smart contract eDSL in Yatima.
+  - Integrating "effectful" features like contract deployment, calling or
+    compatiblity with `pallet_contracts`.
+
+We would plan to propose the above for follow-up work upon satisfactory
+completion of the above deliverables.
+
+### Ecosystem Fit
+
+* Where and how does your project fit into the ecosystem?
+
+Yatima fits into the Substrate ecosystem in a few ways:
+
+1. The runtime can be used like a functional VM that could potentially be targeted by a
+   variety of other functional programming languages, including Plutus, Pact,
+   Simplicity, etc.
+2. The Yatima language itself can be called from within Rust, and potentially
+   can integrate with `ink` as a kind of lightweight functional evaluator or
+   garbage collector for pure expressions, with `ink` or `pallet_contracts`
+   handling calling conventions and effects. This would be similar to the
+   relationship between JS and WASM on the web.
+3. The Yatima typechecker can be embedded in a contract and exposed as an
+   on-chain verifier or prover.
+4. In the future, Yatima could handle effects and contract calls itself,
+   and can become a fully viable smart contract language.
+5. Eventually there could be Yatima-based parachains with entirely different
+   models for how smart contracts should interact.
+
+* Who is your target audience (parachain/dapp/wallet/UI developers, designers, your own user base, some dapp's userbase, yourself)?
+
+Our target audience is smart contract and parachain developers.
+
+* What need(s) does your project meet?
+
+One of the strengths of a heterogeneous blockchain ecosystem like Substrate /
+Polkadot / Kusama ecosystem is the ability for radically different approaches to
+coexist and provide value to one another. Other systems have more of a
+requirement for a single execution environment. So while in the long run we hope
+to execute on our ambition of "better living through FP", even in the short run
+the existence of an interoperable FP option will showcase one of the key
+Substrate innovations in this space.
+
+Additionally, Yatima is designed in a modular way, so that different components
+and dependencies can provide value to other projects, such as `sp_im`, `sp_ipld` libraries,
+or the graph-reduction runtime.
+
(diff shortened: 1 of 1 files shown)


semuelle:
Hi John, thank you for your application. Looks very interesting! We will properly review it as soon as possible.

Noc2 (changes requested):
Thanks for the application. This sounds really interesting, especially your last milestone. Are you aware of https://rs-ipfs.github.io/offchain-ipfs-manual and https://github.com/rs-ipfs/substrate/tree/offchain_ipfs or https://github.com/ipfs-rust/libipld (also see https://github.com/ipfs-rust/libipld/pull/32#issuecomment-620118579)?

Noc2:
Could you also share your previous work for the IOTA and Ethereum foundation? I would also be interested in learning more about your long term plans. Do you potentially want to create a parachain based on this yourself?

johnchandlerburnham:
> Are you aware of https://rs-ipfs.github.io/offchain-ipfs-manual and https://github.com/rs-ipfs/substrate/tree/offchain_ipfs or https://github.com/ipfs-rust/libipld (also see ipfs-rust/libipld#32 (comment))?

Yes, offchain-ipfs is a really cool project, and we actually tried integrating with that and rs-ipfs first. Unfortunately we ran into some issues, and it was unclear how mature and active the development of those projects are. So rather than take on unknown technical debt we decided to fork `libipld` directly for the serialization/deseraliaztion and CID construction functionality we immediately needed for Yatima. This does require that if we want to actually store/distribute Yatima programs over IPFS we have to use an external IPFS node, like `go-ipfs` or `js-ipfs`, which wouldn't be advisable in a Substrate context. So the comment that you linked to seems right to me, the correct long term solution is to have an embeddable ipfs node in Substrate that shares the DB and networking. That's a very worthy engineering effort that we're 100% interested in contributing to.

For the immediate Yatima MVP though, it's not technically necessary, since we keep an internal store of:

```
/// A map of content-ids to defs, with content ids for the def
pub struct Defs {
  pub defs: HashMap<Cid, Def>,
  pub names: HashMap<Name, Cid>,
}
```

Which means you can parse, check and eval Yatima expressions packages locally just like any other language. But yeah, to make Yatima a practical smart contract language on Substrate it's going to be much more efficient to use the Substrate db.

> Could you also share your previous work for the IOTA and Ethereum foundation? 

Sure, so Gabriel Barreto and I previously coauthored the Formality proof language, which was supported by an Ethereum grant. Our work on Formality was an important foundation for Yatima's type theory (though not its implementation or runtime, which are novel), so we wanted to make sure to acknowledge their support. Our DevCon5 presentation on Formality is [here](https://www.youtube.com/watch?v=3XUKQX4nN8w), and our other coauthor Victor Maia is continuing the Formality project [here](https://github.com/uwu-tech/Kind), rebranded to Kind Language.

The IOTA Foundation gave us a grant to develop Yatima and explore integrating it with their WASP node for WASM smart contracts: https://github.com/iotaledger/wasp. Our work on the latter, where we successfully got the language building in WASP-compatible WASM, is [here](https://github.com/yatima-inc/yatima/tree/acs/wasp). We hope to continue that work at some point in the future when WASP stabilizes more.

> I would also be interested in learning more about your long term plans. Do you potentially want to create a parachain based on this yourself?

Absolutely, and we think that would be a great way to both drive adoption of Yatima and give us the resources to continue development. Ideally we'd like to eventually bid for one of the Polkadot parachain slots, but since that's a scarce resources we'd also be fine with a bridge or parathread, or even supporting other chains that want to integrate Yatima.

However, we do believe that a Yatima-specific parachain is an interesting solution architecturally, since there's a possibility of using the sharing in the Yatima runtime on a global-level across an entire block. For example, if I have a transaction that computes some expensive function `f(a)` and you have a transaction that computes the same `f(a)` in the same block, then with sharing we should be able to have `f(a)` computed only once for that block, which will save significant resources. We currently do something similar across package import boundaries (whole program optimization), although this is still rudimentary compared to e.g. [GRIN](https://grin-compiler.github.io/)

Other plans we have include adding green-threads to the language by integrating with the Lunatic VM: https://github.com/lunatic-solutions/lunatic, or exploring frontend development with something based on the https://seed-rs.org/ model. 

The end goal is to have be able to use Yatima for full-stack type-driven development, whether you're writing code on the frontend or backend, on-chain or off-chain.

Noc2 (dismissed):
Thanks for the detailed information. I’m happy to support your project and will share it with the rest of the team.

mmagician:
It's indeed an interesting proposal. I've got a few conceptual questions:

1. I'm having a hard time imagining how using Yatima would work in practice. If it compiles to WASM, would a developer be able to use `pallet-contracts` to upload the WASM blobs there directly, so that the pallet executes these? (so in this scenario, it would be equivalent to a smart contract programming language like `ink!`)

2. You mention an idea of potential integration with `ink!` - so could you call/delegate some logic from within an `ink!` contract to Yatima?

3. Smart contracts are, among other things, a tool for writing & reading on-chain storage. How will you handle that? I guess this is not covered by the scope of the grant?

4. > We will create a Substrate module that will allow for parsing, typechecking and evaluating Yatima language expressions

But it won't evaluate these expressions **directly** on Yatima, but rather execute the WASM that Yatima compiled to, right? Which would clash with the idea from point 1 -> so I guess I'm missing something here.

johnchandlerburnham:
@mmagician 

> 1. ... If it compiles to WASM, would a developer be able to use pallet-contracts to upload the WASM blobs there directly, so that the pallet executes these?

There's a distinction between running Yatima on WASM via the compiled Yatima runtime vs directly compiling a Yatima program to WASM binary. Currently we do the former, which is loosely analogous to how when you compile Haskell with GHC, your binary has the GHC runtime system which does garbage collection, FFI, concurrency, etc, although the Yatima runtime is much slimmer than the GHC RTS (Our lambda-DAG machine is basically a pointer graph). So you wouldn't want to package the Yatima runtime inside a (runtime + program) WASM blob that gets executed by `pallet-contracts`, rather you'd want to extend pallet-contracts with a library `sp_yatima` that lets it directly interpret a serialized Yatima program. 

The other approach is to compile a Yatima program directly to WASM without any runtime. This can be done with techniques like lambda lifting, but requires statically determining allocation and freeing. There are some interesting projects like [the neut language](https://github.com/veka41/neut) that show how this can be done with dependent types. And of course the Rust borrow checker is the best example of using substructural types for this. Yatima has both dependent and substructural types, so we think producing no-runtime WASM or x86 binaries from Yatima is certainly an exciting future possibility, but that's not what we do now.

Even if we had direct compilation though, we would still want to package the Yatima language core inside Substrate so that we could have the ability to parse and typecheck the language on-chain, which opens up the possibility of having a trustless formal verifier contract.

> 2. You mention an idea of potential integration with ink! - so could you call/delegate some logic from within an ink! contract to Yatima?

This is another benefit of having the language core as a library. We currently have a `yatima!` Rust quasiquoter macro that would you do something like:

```rust
let x: u64 = 1;
let f: Term = yatima!("(lambda x => U64.add 1 x)");
let res: Term = yatima!("($f $x)", f, x).eval();
assert!(res.unwrap(),  2u64)
```

(The current syntax is a little uglier for this, but it wouldn't take too much work to implement something close to the above)

So we can imagine using Yatima as like a kind of  "`Rc` for closures" within something like `ink!` potentially, where `ink!` would handle the interaction between the contract and the rest of the network, but you could write pure functional expressions via yatima.

> 3. Smart contracts are, among other things, a tool for writing & reading on-chain storage. How will you handle that? I guess this is not covered by the scope of the grant?

This is a complicated topic, and I wanted to exclude it from the scope of the current grant because it seemed hard to come up with a great deliverable with the time and resources of the initial Open Grant. But once we have `sp_yatima`, I want to propose a follow-up for this. My initial idea is to use dependent records to model the common  "state + methods" style of smart contracts.  (I put a short sketch of how that might look syntactically in a [gist](https://gist.github.com/johnchandlerburnham/5d2703a5dd92f0e0db8225bb9e3121e8), since this comment is getting kinda long)

> 4. But it won't evaluate these expressions directly on Yatima, ...

I think my answer to question 1 touched on this, but let me know if there's anything I can expand on!

semuelle (dismissed):
Excited for this

applications/yatima.md
```diff
@@ -0,0 +1,439 @@
...
+Furthermore, since Yatima can be implemented in itself, there is a possiblity of
+having on-chain trustless proof verification.
+
+Lastly, Yatima implements content-addressing, allowing all programs and packages
+to be serialized into IPLD objects and shared over the peer-to-peer IPFS network
+(or via libp2p).
+
+#### Deliverables
+
+- A `sp_ipld` crate for interacting with IPFS CID, Multihash and IPLD based on
```
Lederstrumpf, line 62: can you integrate concrete feature specifications for these in the description of milestone 1 below?

applications/yatima.md
```diff
@@ -0,0 +1,439 @@
...
+  [`ipfs-rust/libipld`](https://github.com/ipfs-rust/libipld), but which is
```
Lederstrumpf, line 63: is this a wrapper around `libipld` that avoids use of `std`, or is there more to it?

applications/yatima.md
```diff
@@ -0,0 +1,439 @@
...
+  Substrate pallet compatible. This is a necessary dependency for Yatima Core,
+  but is an independent deliverable that could be useful for many other
+  Substrate projects.
+
+- A `sp_im` crate for creating immutable datastructures, based on
+  [`bodil/im-rs`](https://github.com/bodil/im-rs), but which is
```
Lederstrumpf, line 69: ditto above comment re. `libipld`

applications/yatima.md
```diff
@@ -0,0 +1,439 @@
...
+  Substrate pallet compatible. This is a necessary dependency for Yatima Core,
+  but is an independent deliverable that could be useful for many other
+  Substrate projects.
+
+- A `sp_yatima` crate integrated with `sp_ipld`, `sp_im` and `sp_std` that
```
Lederstrumpf, line 74: for demonstration purposes, could you include rudimentary integration with `sp-storage`? Just read-level operations. I think that would be useful for a follow-up grant motivation.

Lederstrumpf (changes requested):

applications/yatima.md
johnchandlerburnham: Sure, what concrete feature specifications should we add? Milestone 1 lists:

"serializing and deserializing IPLD objects and computing IPFS CIDs"

Which is the essential functionality, but we can also add a description of the dependencies:

- bytecursor (no_std replacement for Rust Read/Write traits, WIP repo is here: https://github.com/yatima-inc/bytecursor)
- sp-cid (no_std port of https://github.com/multiformats/rust-cid, WIP repo is here: https://github.com/yatima-inc/sp-cid
- sp-multihash (no_std port of https://github.com/multiformats/rust-multihash, WIP repo is here: https://github.com/yatima-inc/sp-multihash)

Also we should clarify that we were only planning on having `sp-ipld` support dag-cbor, but I believe we could prettily add dag-json support (and a json textual format for both) if this is desirable.

applications/yatima.md
johnchandlerburnham: It actually requires changing most of the ipld/ipfs/multiformats dependency tree, primarily to replace usage of Read/Write, there are also some additional places where `sp_std` isn't quite a drop-in replacement, such as needing `alloc` for `String`, etc. These changes have to then be documented and tested for compatibility with `go-ipfs`, which is where the bulk of the work is. We are using `quickcheck` for this, and have already discovered a few corner cases where `libipld` does not compute the same CID as `go-ipfs` does. We plan to upstream our fix for that case, and any other fixes we discover to `libipld` as well.

applications/yatima.md
johnchandlerburnham: The changes to make `im` `no_std` compatible are also extensive, and primarily consist of simplifying the library so as to remove the `im_rc` non-threadsafe features, refactoring the test-suite, and removing HashMaps (due to the Substrate's restriction https://substrate.dev/docs/en/knowledgebase/advanced/no-hash-collections). Since the parent repository `im` appears potentially unmaintained: https://github.com/bodil/im-rs/issues/185, we're preparing to publish and maintain `sp-im` as an independent library.

applications/yatima.md
johnchandlerburnham: I agree that integrating with `sp-storage` would be useful, but integrating read operations directly into the language is not actually much less work than integrating both read and write operations. In both cases we're effectfully interacting with an external DB, and have to correctly compute key values and correctly model StorageValue return types, both of which seem non-trivial. We have been considering two different approaches to this, which may not be mutually exclusive:

1. Creating a custom SubstrateIO type in Yatima which mirrors the APIs found in https://substrate.dev/rustdocs/v3.0.0/frame_support/storage/trait.StorageMap.html#required-methods and https://substrate.dev/rustdocs/v3.0.0/frame_support/storage/trait.StorageValue.html.
2. Extend yatima_core with first-class-modules primitives (dependently typed maps) and use the modules as a kind of frontend to the StorageMaps without requiring everything to be mapped to monadic IO.

Both cases involve extending our runtime DAG machine with primitives that actually perform the right Substrate calls, so that's probably where we'll start. Not 100% sure it's really in scope as part of the current proposal, which is essentially "get the pure language working", but it's definitely Milestone 1 on our follow-up.

applications/yatima.md
johnchandlerburnham: I've gone ahead and added these to Milestone 1, since the dependencies are implicit anyway, and `dag-json` is a great additional feature to have.

Noc2 (approved):

applications/yatima.md
Lederstrumpf: Thanks for the clarification. On a high-level, I'd prefer the second approach, but let's cross that bridge once we get there.

Lederstrumpf (approved):
Thanks for the detailed clarifications @johnchandlerburnham!

Your application is excellent and I really look forward to the fruits of your work :)

semuelle (approved):

johnchandlerburnham:
Thanks everyone!

PR #510, opened 2021-07-20
Update faq.md

semuelle:


Noc2:
We should improve the part about the token sale at some point. Otherwise it looks good.

PR #511, opened 2021-07-20
Update pull_request_template.md

alxs:
@Noc2 I don't think this needs much updating. If private finances/funding above limits, they can leave the second checkbox unchecked, if fiat the third one.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,5 +1,15 @@
 # Grant Application
 
+## Abstract
+
+> Provide a brief description of your project here summarising key points (1-2 paragraphs).
+>
+> If your application is a follow-up to a previous one, please mention which one in the first line of the abstract and include a link to previous pull requests if applicable.
+>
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -1,9 +1,9 @@
 # Grant Application Checklist
```
alxs, line 2: Although also I quite like this. Maybe we could add it to all applications.

```suggestion

## Abstract

> Provide a brief description of your project here summarising key points (1-2 paragraphs).
>
> If your application is a follow-up to a previous one, please mention which one in the first line of the abstract and include a link to previous pull requests if applicable.
>
> The details should be in the application document that is being added to the repository via this pull request.

```

.github/pull_request_template.md
```diff
@@ -1,9 +1,9 @@
...
```
alxs, line 1:
```suggestion
# Grant Application
```

PR #519, opened 2021-07-22
Delete commit_count.yml

alxs:
This is just a minor cosmetic issue in case teams have a large number of commits. In which case the bot comment further pollutes the thread


mmagician:
I think it's not so much to ask for a clean PR. It is also mentioned when creating a new PR. But I won't defend my ground at all costs here :)

alxs:
Fully agree, I'm only arguing that the bot comment doesn't necessarily help the cause. I would ask manually in case it's really excessive or otherwise live with it (which is the way we've been doing it since forever, seeing as the action has been broken for some time).

Noc2 (approved):
Since it’s broken, I guess we can merge it.

mmagician:
Good point @Noc2

PR #522, opened 2021-07-23
Append data to Google Sheet instead of getting last row

alxs:
This is working nicely for me. @mmagician I don't know what the problem was originally, but maybe it's been fixed?

If this PR is merged I'll also delete https://github.com/w3f/gsheet.action


Noc2:
Let’s give it a try.

PR #538, opened 2021-07-28
add hazelword.md

ydong3:
Project Name：Hazelword 
add add hazelword.md forW3F Grant Proposal

### DIFF
--- a/applications/hazelword.md
+++ b/applications/hazelword.md
@@ -1,0 +1,189 @@
+# W3F Grant Proposal
+
+* **Project Name:** Hazelword
+* **Team Name:** Hazelword team (e.g. Duo)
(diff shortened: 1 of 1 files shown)


Noc2 (changes requested):
Thanks for the grant application. However, we usually don’t support the development of solidity smart contracts via our grant program. On the one hand we are no experts when it comes to solidity, on the other hand there is a high chance that a similar contract already exists. Furthermore your deliverables are currently really unspecific. For example, what exactly do you mean with “TEE module”. The deliverables and the specification are basically the requirements of our contract and need to have enough details so that we are able to evaluate these deliverables.

semuelle:
@ydong3, @ywdlucking needs to sign the CLA, because they are the one that committed the file to your repo.

ydong3:
Due to our reasons, it cannot be updated at #538, so keep to #551

PR #542, opened 2021-07-29
Uncollateralized Stablecoin RFP

Noc2:
As promised I created an RFP for an non-collateralized stable coin. Let me know in case you know of such a stablecoin that already works ;-)

### DIFF
--- a/rfps/uncollateralized-stablecoin.md
+++ b/rfps/uncollateralized-stablecoin.md
@@ -27,13 +27,13 @@ The milestones below are just an initial draft. The milestones can be structured
 
 * **Total Estimated Duration:** 2 month
 * **Full-time equivalent (FTE):**  1
-* **Total Costs:** 20k
+* **Total Costs:** 30k
 
 ### Milestone 1 - Research & Design 
 
-* **Estimated Duration:** Duration of milestone 1 
-* **FTE:**  Amount of time (in days) required for a single person to complete this milestone
-* **Costs:** Amount of Payment in USD for milestone 1
+* **Estimated Duration:** 1 month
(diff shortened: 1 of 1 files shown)


rfps/uncollateralized-stablecoin.md
```diff
@@ -0,0 +1,56 @@
...
+* **Estimated Duration:** Duration of milestone 1 
+* **FTE:**  Amount of time (in days) required for a single person to complete this milestone
+* **Costs:** Amount of Payment in USD for milestone 1
+
+
+| Number | Deliverable | Specification | 
+| ------------- | ------------- | ------------- |
+| 1. | Overview of existing solutions | Create an overview of existing and previous uncollateralized stablecoin solutions and why they failed. |  
+| 2. | Increasing Supply | Develop a mechanism to deal with an increasing demand for the cryptocurrency and analyses potential attack vectors | 
+| 3. | Decreasing Supply | Develop a mechanism to deal with an decreasing demand for the cryptocurrency and analyses potential attack vectors|
```
alxs, line 43: I feel like whether it's worth it to go ahead and do milestone 2 at all may depend too much on this being done right, and like the likelihood of a team delivering something that looks reasonable at first glance but is inherently broken for this part is too high to make the implementation part of the same grant.

This may be a good opportunity to have a research-only rfp/grant.

Noc2: Makes sense. I will update it

Noc2:
I’ll go ahead and merge it. We can still add additional details or change it later. 3x fireworks counts as three approvals ;-)

PR #544, opened 2021-07-30
OpenSquare off-chain voting platform

wliyongfeng:
### Project Abstract

In short, you can see this proposed platform as [snapshot](https://snapshot.org) in the polkadot ecosystem.

### Application Checklist

- [x] The total funding amount of the project is below USD $30k for first-time grant applications and $100k for follow-up ones.

### DIFF
--- a/applications/OpenSquare-offchain-voting.md
+++ b/applications/OpenSquare-offchain-voting.md
@@ -1,0 +1,114 @@
+# W3F Grant Proposal
+
+* **Project Name:** OpenSquare off-chain voting for Polkadot ecosystem
+* **Team Name:** OpenSquare
+* **Payment Address:** 0x4905083abdD13bd95345A871701Fd0b08AbD46d1 (USDT)
+
+## Project Overview :page_facing_up:
+
+### Overview
+
+In short, you can see this proposed platform as [snapshot](https://snapshot.org) in the polkadot ecosystem.
+
+Though Polkadot/Kusama and most other Substrate based chains have on-chain governance and runtime can be easily upgraded with democracy. We think this platform still make sense with following reasons:
+- Compared to the formal on-chain votes, this platform provides not so formal off-chain signed polls which may help token holders and community members engage **more** in the ecosystem building.
+- Compared to Polkassembly polls on post:
+  - users have to sign their votes and the signed data will be stored on IPFS
+  - users don’t have to do any registration. 
+  - Different strategies can also be provided to calculate the final result, rather than simply count the vote numbers.
+- Not all on-chain systems with assets on Substrate based chains can upgrade with democracy, like assets issued on Statemine, ERC-20 assets which may be issued on Moonriver, Karura, Shiden, etc.
+
+### Project Details
+Off-chain votes will be an important part of OpenSquare collaboration platform. Other planned collaboration forms include bounties, payment QA, short term employment, etc. Key components in this proposal include:
+- Predefined spaces where users can create proposals. It will definitely include Kusama and Polkadot, current para-chains with very high possibility. 
+- Proposal list in one space where users can view the closed or ongoing proposals.
+- Proposal detail page where users can view the detail and the signed votes.
+- A proposal discussion arena where users submit the signed messages and here we expect chaos.
+- An authoring page where users can create proposals, and set the corresponding snapshot height, start and end date(height).
+- Archive nodes interaction with which we read users’ balances on the corresponding snapshot block height.
+- A backed server to interact with the node(Polkadot, Kusama, etc), store the organized data to DB, handle IPFS storage, provide APIs to get data.
+
+![os_grant_voting](https://user-images.githubusercontent.com/2264908/127607269-5d54f0b0-d8b8-48f1-9c7a-fc8c205bc645.png)
+
+Some implementation notes:
+- We have to call polkadot.js extension to sign the voting/poll, and some discussion messages.
+- About voting power, since Polkadot/Kusama have proxy accounts, so we have to support proxy account vote on behalf of the original one.
+- In this proposal, we will implement `balance of account` and `quadratic balance of account` strategies for Polkadot and Kusama. It means if Alice's balance is 100 at one chain height, her voting power will be 100 and 10 by these 2 strategies.  
+- We have a partnership with [Crust](https://crust.network/), and we many store data to IPFS through [decoo](https://decoo.io/) that crust granted. 
+
+### Ecosystem Fit
+- Providing off-chain voting/polls to help token holders/community members engage more in ecosystem building.
+- Flexible strategies help produce different voting/poll results which bring us different opinions from the on-chain tallying methods.
+- [snapshot](https://snapshot.org) is popular for Ethereum stack projects, mainly for governance, and currently we didn't see similar projects in Polkadot ecosystem.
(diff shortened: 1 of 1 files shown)


Noc2:
Just to add. This looks really good, but could you update the milestone deliveries according to our template: https://github.com/w3f/Grants-Program/blob/master/applications/application-template.md#milestone-1-example--implement-substrate-modules Also adding more technical information about the license, testing, programming language, polkadot.js extension etc. Apart from that I think user stories are already good deliveries. Maybe also see milestone 1 of this application: https://github.com/w3f/Grants-Program/blob/master/applications/bright_treasury.md#milestone-1--idea-creating--proposal-submission--in-app-logins

wliyongfeng:
@Noc2 Thanks for the feedback. Have correct the deliveries format. 

I tried to give more technical implementation details in the [deliveries table code cell](https://github.com/w3f/Grants-Program/pull/544/files#diff-36130f52c98ebb021fb880e5fe73e261a0397914674ea55d4c375e81f018dacdR99) and [implementation notes](https://github.com/w3f/Grants-Program/pull/544/files#diff-36130f52c98ebb021fb880e5fe73e261a0397914674ea55d4c375e81f018dacdR33). But still not very sure with the 'a higher level of technical detai', please tell me if they are still not enough.

alxs:
Just a note in case there's any more feedback: no need to force-push any more. This is just for the initial PR, subsequent commits are useful as they allow us to see the changes.

Noc2 (approved):
Thanks for the update and sorry for the late reply. I’m happy to go ahead with it. But feel free to remove Deployment & Promotion. That’s something that we usually don’t support via grants, but obviously it’s great if you do it anyway.

wliyongfeng:
> Thanks for the update and sorry for the late reply. I’m happy to go ahead with it. But feel free to remove Deployment & Promotion. That’s something that we usually don’t support via grants, but obviously it’s great if you do it anyway.

Removed and for sure we will deploy it to collaborate with communities and for your better future review.

Noc2 (approved):

alxs (approved):
@wliyongfeng I agree, this actually looks very good. I don't remember exactly what was missing the first time, but perhaps I was a bit trigger-happy with the 'details-missing' label. Looking forward to the results!

mmagician (commented):
To be honest, I don't clearly see how having another platform next to Polkassembly would help the ecosystem, especially that you'd get fragmentation of the user base -> results from off-chain voting would be less representative of the real vote.
I get the feeling that you're trying to differentiate from Polkassembly just "for the sake of it":
- no for registration
- yes for signatures

In the end these are two different mechanisms towards having "credible" votes and you haven't provided justification for why one is superior to the other.

While it's true that we generally via Grants Program we support multiple implementations for the same functionality, these usually doesn't have an impact on the global outcome (e.g. the staking rewards collector will be used by a single user, and the ecosystem doesn't suffer if one user goes with implementation A and another user with implementation B). In this case however, my assumption is that a community member will either cast their votes on platform A or B, not both (tracking all the proposals happening is already considerable amount of work, let alone tracking/commenting on multiple platforms).

All that said...it would be great not to rely on a single platform in case the development there ceases. But since polkassembly is open-sourced anyway, that risk is greatly reduced.

I'm still on the fence with your proposal. Have you perhaps considered proposing new features to be merged into [polkassemby](https://github.com/paritytech/polkassembly) instead?

wliyongfeng:
Hi, @mmagician .

Understand, and we will be happy if polkassembly can have rich of features that we expected. We received same concern when we develop [dotreasury](https://www.dotreasury.com) and [Statescan](https://www.statescan.io/statemine). I think we don't have to discuss too much about questions like, why subscan while there are polkascan, why google while there are yahoo. Yes, we can add these proposed features to polkassembly, but I'm very worried and hesitated about following truth:
- I have to setup polkassembly and understand every code to contribute my expected features
- Every of my PRs have to pass the review. It main include some rules like code style, test coverage requirement, etc
- We have to use some tech stack that we don't like. Actually we checked the code of polkassembly and it's the truth that we're not familar with some tools or libraries

For the reasons above, contributing to polkassembly directly will cost at least double time/cost. And we believe many contributors will quit if the requirement is contributing to similar projects.

What we expect is to deliver the features with least time and lowest cost. Then let's see what happens, maybe let some products merge to one if possible.

About the fragmentation, we don't think polkassembly contains all the features we proposed and we don't think that we don't have to do it since polkassembly may have it in the future. And we can let this platform down or just redirect to polkassembly, when polkassembly support it.


> representative of the real vote

We don't expect this off-chain voting/polls will take place of on-chain voting. We expect community members engate more in ecosystem building, while I think the reason for many community members not vote on-chain is just avoid gas. And it may also provide ways of voting for Statemine and ERC-20 assets easily.

mmagician:
> I have to setup polkassembly and understand every code to contribute my expected features
> Every of my PRs have to pass the review. It main include some rules like code style, test coverage requirement, etc

Well, these are expected when contributing to *any* codebase. I would strongly argue that the second point is an advantage rather than a disadvantage, since it enforces good coding practices, which leads to cleaner, and especially safer, code.

> We have to use some tech stack that we don't like. Actually we checked the code of polkassembly and it's the truth that we're not familar with some tools or libraries

Again, this is going to be true for any codebase. You will find that the tech stack *you* use is unfamiliar to other community members and that they prefer to work with different tools. Consider this: if every time that I encounter a new project that I'd like to contribute to, I decide to re-build most functionality from scratch with a different tech stack that I'm more comfortable with, then open source won't get anywhere. I acknowledge there some diversity is needed, but I strongly disagree with the reasons you provided above for diversification.

wliyongfeng:
> I would strongly argue that the second point is an advantage rather than a disadvantage

Sure for one codebase robustness, but maybe not for fast and low cost features delivery. What we expect is to provide a off-chain solution like snapshot quickly. 

Honestly I don't care about the code quality of polkassembly, and don't want to take the trouble to add features to polkassembly. And as I said, after this delivery, polkassembly developers can copy the features to their repo, and we can even make our voting off if you think it should be on polkassembly officially.

> I strongly disagree with the reasons you provided above for diversification.

The main reason is the time & cost. And most of our code is open source too.

NhlanhlaHasane (approved):

semuelle (commented):
@wliyongfeng, my only question is: if you jump through the hoop of storing the voting data on IPFS, will you make the location & format publicly available? Or what is the benefit of using IPFS?

wliyongfeng:
@semuelle 
Thanks for the comment. Sure, we try to make it public and not be private collection. Benefits of using IFPS:
- User can get the IPFS link of their voting/proposal data immediately. It will be a proof that we can not just keep the data, not show it and affect the result.
- In case of data loss and we can recover the data in a better way if our server with database down.


@mmagician 
Sorry if my comments are still not convictive, but still hope to get your support. Supporting statemine assets will be our next plans, and I think it will make more sense while they may have no any on-chain governance. We tried to narrow the scope to build some basic modules first and for better review.

We will be very glad to merge some of features to polkassembly or other platforms if they verified works, and of course, at a better time.

semuelle (approved):
My comment regarding IPFS was more along the line of others being able to build alternative UIs or features around the data, but either way, I'm happy to support this.

PR #559, opened 2021-08-04
Unify welcome comment + Google Sheet workflows

alxs:
See https://github.com/alxs/Grants-Program/actions/runs/1097707799

Both jobs would run if and only if a file has been added to `applications`. I'd say there's no need to check whether the author is a member of the grants team or the application is labelled 'amendment'.


Noc2:
K, let’s test it in case you haven’t already.

PR #560, opened 2021-08-04
Update badge guidelines

alxs:
- Style issues
- Technically the committee doesn't accept milestones, we do
- We probably at least don't want to _encourage_ teams to think of it as a medal or trophy
- Clarify what constitutes a project-specific vs teams/project's landing page
- Questions on the matter are probably better addressed to grantsPR


semuelle:
nice

PR #561, opened 2021-08-05
Delete README_CN.md

alxs:
Seems no one wants to update it (so probably it'd also be a nightmare to maintain up to date) and this version is now heavily outdated.

It's also getting close to 0 traffic.


mmagician (approved):

Noc2:
Fine with me. Maybe we can give it another shot, once the level system is merged.

PR #438, opened 2021-06-03
Apply Grant Application for Gateway

hskang9:
# Grant Application Checklist

- [x] The [application template](https://github.com/w3f/Open-Grants-Program/blob/master/applications/application-template.md) has been copied, renamed ( `project_name.md`) and updated.
- [x] A BTC or Ethereum (DAI) address for the payment of the milestones is provided inside the application.  
- [x] I have read and acknowledged the [Terms and Conditions](https://github.com/w3f/Open-Grants-Program/blob/master/src/T&Cs.md).
- [x] The software delivered for this grant will be released under an open-source license specified in the application.
- [x] The total funding amount of the project is below USD $30k for initial grant applications and $100k for follow-up grants.
- [x] The initial PR contains only one commit (squash if needed before submitting your PR).
- [x] The grant will only be announced once the first milestone [has been accepted](https://github.com/w3f/Grant-Milestone-Delivery).

PR #562, opened 2021-08-05
Update polkadot_stack.md

alxs:
Not sure if this makes sense as a category, I feel it deserves its own outside of DeFi. Also we had another one of these, no?


Noc2 (approved):
Make sense. I think the correct term for this is now HyFi (hybrid finance) = Cefi + Defi. But most people probably don’t know about it and I generally don’t like abbreviations. Most of the time it’s just to sound more intelligent ;-)

semuelle:
TradFi (Integration)?

alxs:
+1 for HyFi, though I agree there's no need to get fancy with the abbreviations.

Things like [this](https://www.prnewswire.com/news-releases/current-acala-announce-partnership-establish-new-category-of-hybrid-finance-301293713.html) and [this](https://github.com/w3f/Grants-Program/pull/438) would also belong in this category, but is there any other existing project we could add?

PR #567, opened 2021-08-05
Added Fei Protocol whitepaper to stablecoin RFP resources

jakehemmerle:
Fei Protocol creates a stablecoin that uses direct incentives and protocol owned value to peg to a price. It's capital efficient, decentralized, and doesn't rely on an over-collateralization mechanism.

Personally, I think the project is pretty promising.

https://coinmarketcap.com/currencies/fei-protocol/


Noc2:
Thanks for adding these links to the RFP. I will also add Terra (LUNA), which I think is the most promising after looking into it a little bit more.

PR #582, opened 2021-08-18
Move levels section before process

alxs:
I think it makes more sense for teams to learn about the levels _before_ they apply ;)


mmagician (approved):

Noc2:
Makes sense

PR #593, opened 2021-08-27
feat: add SubGame Network m2

xtony77:
### Project Abstract

In [Milestone 1](https://github.com/w3f/Grants-Program/blob/master/applications/SubGame_Network.md), we created the first game module, and then we will add more game modules for users to apply. Now, in the Milestone 2 of SubGame, we will launch other game modules in succession. In order to display and explain these game modules and circulate them, we will build a pallet module lease platform where other users can lease. The way of leasing is to obtain the right to use the module according to his needs, so that other developers can easily develop games through the platform. Using the advantages of blockchain decentralization, compared with real leasing, to avoid the risk of performance, as well as the full traceability of leasing, to ensure the authenticity and non-tampering of data, which will upgrade the virtual currency ecology to a new level.

- [ ] **Level 2**:  Up to $50,000, 3 approvals

### Application Checklist

### DIFF
--- a/applications/SubGame_Network_m2.md
+++ b/applications/SubGame_Network_m2.md
@@ -199,13 +199,13 @@ No Legal Entity
 ### Overview
 * **Total Estimated Duration:** 8 weeks
 * **Full-time equivalent (FTE):** 4 FTE  
-* **Total Costs:** 15,000 USD
+* **Total Costs:** 10,000 USD
 
 ### Milestone 2 : Build a module leasing platform
 
 * **Estimated duration:** 8 weeks
 * **FTE:** 4 FTE 
-* **Costs:** 15,000 USD
+* **Costs:** 10,000 USD
 
 | Number | Deliverable | Specification |
 | -----: | ----------- | ------------- |
--- a/applications/SubGame_Network_m2.md
+++ b/applications/SubGame_Network_m2.md
@@ -199,13 +199,13 @@ No Legal Entity
 ### Overview
 * **Total Estimated Duration:** 8 weeks
 * **Full-time equivalent (FTE):** 4 FTE  
-* **Total Costs:** 0.6 BTC 
+* **Total Costs:** 15,000 USD
 
 ### Milestone 2 : Build a module leasing platform
 
 * **Estimated duration:** 8 weeks
 * **FTE:** 4 FTE 
-* **Costs:** 0.6 BTC
+* **Costs:** 15,000 USD
 
 | Number | Deliverable | Specification |
 | -----: | ----------- | ------------- |
--- a/applications/SubGame_Network_m2.md
+++ b/applications/SubGame_Network_m2.md
@@ -1,0 +1,240 @@
+# Open Grant Proposal
+
+* **Project Name:** SubGame Network m2
+* **Team Name:** [SubGame-Network](https://github.com/SubGame-Network)
+* **Payment Address:** 3CK6HN4y2ZEKX7JMfdegds7B7GrUSmv65c
+
+## Project Overview :page_facing_up: 
+### Overview
+
+![](https://i.imgur.com/UhLDmyL.png)
+
+* **Introduction**
+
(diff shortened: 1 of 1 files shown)


mmagician:
Please note that the total cost should be specified in USD - bitcoin payment is possible, but the exchange rate will only be determined at the time of payment.

xtony77:
@mmagician The total cost has been modified to USD.

mmagician:
Thanks. I'm glad to see that you're continuing to work on your project. Could you please explain what happens after a user successfully completes the lease step? What does it mean, on a technical level, that a user has access to a module for a limited time? They have permission to use the source code in their own chain?
Or rather, does the SubGame team deploy each module into your runtime, and then allow the users to call methods on that pallet? (in which case, would you keep some kind of a whitelist of users?). Perhaps you could illustrate this all with an example?

xtony77:
1. First, there is the question of time constraints, the permissions will disappear after the subscription period, and you need to stake to renew. Since our lease platform is a subscription system, it is not free. We will develop some game pallets and put them on the lease platform. Other users can subscribe to our pallet through stake and call the pallet on our chain. Compared with their own Development, leasing on our leasing platform can reduce a lot of development times.
2.  Then, after successfully leasing, the user will go to the following page, which supports the Markdown format, so the developer will directly attach the usage method of the pallet to the top, include how to call methods on that pallet.

    e.g.
    
    
3. Finally, because it is a subscription system, users need to stake the subscription before they can use it. Of course, the pallet is open source, and any user can move it to their own chain, but our lease platform saves users a lot of trouble in setting up infrastructure.

mmagician:
I see, thanks for the answers, that all makes sense. Still, I wonder - the game developer needs to agree to a fixed set of functions inside the game that you deployed with your pallet, and if they want to get updates, they need to rely on you to upgrade the runtime. And I imagine it would only make sense for *one* such developer to use this pallet (otherwise the same game would get deployed twice - not much benefit for neither the users nor developers probably). And why doesn't the pallet creator deploy the game themselves, rather than leasing it out - assuming that the game makes brings some profit, I would think it's in the interest of creator to benefit first, no?

xtony77:
1. I know what you mean. First of all, we lease a pallet. The pallet is a specific function (such as drawing/synthesis/combat... etc.) rather than the entire game, so our expected usage scenario is that the user will lease and need to build The function of own game. Therefore, the main game logic will be packaged by other game developers, but some functions will use our pallet. The degree of use depends on other game developers, so the same game will not appear multiple times.
2. Then, of course we will deploy our own games, but the lease is still the pallet, not the entire game. The reason for this is to maintain the diversity of the game. If it is to lease a deployed game, the same game will appear multiple times like you said. Finally, if it is operated in the form of a lease pallet, because the pallet is a single function, although users may rely on our upgrade, the impact is not significant.

mmagician:
Right, makes sense. I have a really hard time imagining this model being adopted by developers (compared to the devs building and deploying their own modules), despite all the arguments you've provided. I do think the general idea of leasing out modules is very interesting though and is something that could be re-used by the wider community, I hope - so overall, I'm on the fence here.
Would it perhaps be possible for you to reduce the cost such that you fall under Level 1 grant (i.e. to $10,000)? Let me know if this option is feasible, I think this way it would be much easier to get your application through with the committee (I'll be already happy to support, so then you would just need 1 extra approval).

xtony77:
We understand what you mean. The team also hopes to be approved. In terms of price, our team has some flexibility. However, Level 1 is a Individuals plan. We still hope to pass the review as Level 2, thanks!!

mmagician:
It is not strictly necessary to be a solo developer to apply for Level 1. Mostly it's about the price - the lower the price, the quicker it should be to get the necessary approvals.

xtony77:
Okay, we understand that our team has revised the price and handed it to you for approval, thanks!

mmagician (approved):

Noc2 (approved):

PR #117, opened 2020-11-23
Manta: A Privacy Preserving Decentralized Exchange

stechu:
# Grant Application Checklist

- [x] The [application-template.md](https://github.com/w3f/Open-Grants-Program/blob/master/applications/application-template.md) has been copied, renamed ( "project_name.md") and updated.
- [x] A BTC address for the payment of the milestones is provided inside the application.  
- [x] The software of the project will be released under the Apache license version 2.0 as specified in the [terms and conditions](https://gist.github.com/Noc2/75bc58e8ce9b5d419ff883b0cf2b8c19).
- [x] The total funding amount of the project is below CHF 30k at the time of submission.

PR #606, opened 2021-09-08
update manta milestones

stechu:
#117

### DIFF
--- a/applications/manta_network.md
+++ b/applications/manta_network.md
@@ -171,9 +171,9 @@ The Manta substrate pallets includes the ledger state implementation, the transa
 
 ### Milestone 2 — Manta DAP Wallet Protocol and XCM Integration
 
-* **Estimated Duration:** 7 weeks (Estimated Delivery Date: Oct. 30)
+* **Estimated Duration:** 7 weeks (Estimated Delivery Date: Nov. 15)
 * **FTE:**  7
-* **Costs:** 2 BTC
+* **Costs:** 0.5 BTC
 
 There are two major new deliverable in Milestone 2:
 1. Support minting Parachain asset via XCM. This requires the following:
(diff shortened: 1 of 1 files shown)


stechu:
@alxs  we update manta milestone 2 here. Since the development of milestone 1, we obtain a much better understanding, both in terms of the scope and in terms of product features needed to be built.

This updated milestone contains three major additions:

1. Include the wallet protocol and implementation  into the scope: we realized that we need a good wallet protocol to provide the user the same abstraction they used to (Account) rather than the UTXO model mantaPay use for privacy.
2. Include `manta-signer`, the manta native ZKP generator into the scope. We find the pure browser WASM performance is far from a good user experience needs. Thus, we decide to develop a Win/Mac signer to make the transactions sending blazingly fast.
3. Include a dedicated and polished DApp. This is built by React.js and helps user to perform private transactions.

alxs:
Hey @stechu, thanks for the update. A few comments:

- Milestones now need to be denominated in USD. If you want to update the price, please quote it in USD. Furthermore, we're very unlikely to approve a single milestone for ~90k. Given the current delay, I would rather suggest you submit whatever work you've done until now since milestone 1 or only some of the new deliverables, leave the price untouched and then apply for a follow-up grant.
- Most parachain probably have to deal with abstracting currencies, I think you're unlikely to have to implement this yourselves?
- I'm not 100% sure here, but we have a policy to only fund software that doesn't rely on closed-source components to work, which could very well mean we can't fund the Windows/Mac tool. Personally, I'd also be much happier to fund efforts toward improving ZKP performance in WASM.
- Any assets you're actively planning to support? Would you have to coordinate with each individual project for this? (and is XCMP ready?)

stechu:
> Hey @stechu, thanks for the update. A few comments:
> 
> * Milestones now need to be denominated in USD. If you want to update the price, please quote it in USD. Furthermore, we're very unlikely to approve a single milestone for ~90k. Given the current delay, I would rather suggest you submit whatever work you've done until now since milestone 1 or only some of the new deliverables, leave the price untouched and then apply for a follow-up grant.
> * Most parachain probably have to deal with abstracting currencies, I think you're unlikely to have to implement this yourselves?
> * I'm not 100% sure here, but we have a policy to only fund software that doesn't rely on closed-source components to work, which could very well mean we can't fund the Windows/Mac tool. Personally, I'd also be much happier to fund efforts toward improving ZKP performance in WASM.
> * Any assets you're actively planning to support? Would you have to coordinate with each individual project for this? (and is XCMP ready?)

Sounds like a plan. And thanks a lot for the review.
1. We are planning to support DOT and ACA/aUSD as the first batch of assets. 
2. Keep the original price untouched makes perfect since if signer is not a part of the scope
3. We do have plan to improve ZKP circuit and WASM speed. The issue here is that it is a bit research problem so the timeline is a bit uncertain.

alxs:
That means no plans for Calamari? I just saw that the client/signer is part of the original scope, couldn't you just release a Linux version as well?

semuelle:
@stechu, are you still planning on updating the application? Otherwise we would close this PR.

alxs:
@semuelle while I don't think the amendment is strictly required here, the team should certainly submit soon since the milestone is now delayed by 7 months. I wouldn't just close the PR in case the team needs it in order to submit, but if it's unsuccessful (or we don't hear from them soon), maybe we should terminate instead.

@stechu I'd also like to ask you to please add back the Dockerfile, which makes testing easier for us, and the benchmarks which I think make a lot of sense especially on the prover side. Besides, you can probably build on the excellent ones you submitted for M1.

alxs:
@stechu it's now been 3 weeks since your last update and as I mentioned in my last comment, this grant is already delayed by 7 months. If we don't hear from you within the next week, I'm afraid we will have to terminate the agreement.

PR #627, opened 2021-10-01
Create logion_wallet.md

elisadba:
### Project Abstract

> Please provide a brief description of your project here summarising key points (1-2 paragraphs).
>
> If your application is a follow-up to a previous grant, please mention which one in the first line of the abstract and include a link to previous pull requests if applicable.

- [ ] **Level 2**:  Up to $50,000, 3 approvals

### Application Checklist

- [ ] A BTC or Ethereum (DAI/USDT) address for the payment of the milestones is provided inside the application.
- [ ] I have read and acknowledged the [terms and conditions](https://github.com/w3f/Grants-Program/blob/master/docs/T&Cs.md).
- [ ] The software delivered for this grant will be released under an open-source license specified in the application.
- [ ] The initial PR contains only one commit (squash and force-push if needed).
- [ ] The grant will only be announced once the first milestone [has been accepted](https://github.com/w3f/Grant-Milestone-Delivery#process).


gdethier:
@mmagician the off-chain part of our project actually looks more important than it is. The backend will not run on our servers but in a decentralized way beside each node. The purpose of the backend instances is to store/process the confidential/sensitive data (e.g. user identity) that will not make it on-chain. While those data are necessary for what we do (business-wise I mean), the core of all our processes is implemented on-chain.

I just realised that the figure part of the application and actually showing this is missing. I fix that now.

mmagician:
Sounds good. Do you intend to package your backend application together with your node? Maybe it's worth taking a look at how another grantee handled something similar:
https://github.com/Evanesco-Labs/substrate

You should be able to see what they adapted the `client` in their commits to run another service alongside the node.

gdethier:
No, we do not intend to package our backend with the node (it is a standalone TypeScript service interacting with the node using Polkadot JS). But thanks anyway for the interesting pointer, I'll look into that.

mmagician:
Then otherwise how will you ensure decentralisation? Do you have any incentives for service providers who also run the backend?

gdethier:
In terms of software, logion's Substrate node and the backend are two separate components. However, on each node of the network, there will be one Substrate node service *and* one backend service running. In other words, logion "service providers" (their nature is quite specific, see the grant request) will always run a Substrate node *and* a backend service on each of their logion nodes. Therefore, each node being self-contained and independent, we achieve decentralisation. So if by "packaged together" you mean that the node and the backend actually run on the same server, then the answer is "yes".

gdethier:
@mmagician Sorry, I realize I did not mention you in my previous reply so you probably missed it. Does it make more sense now?

Noc2 (changes requested):
Thanks for the application. Looks interesting. I especially like the fact that you want to leverage the [recovery pallet](https://github.com/paritytech/substrate/blob/master/frame/recovery/src/lib.rs). See this [RFP](https://github.com/w3f/Grants-Program/blob/master/rfps/social-recovery-wallet.md). Could you maybe specify the deliveries as part of the milestone a little bit better (Specification section)? For example, the function and the programming language. This way it will be easier to evaluate the project. Do you already have some initial mock-up or design regarding the webapp, which you can share?

semuelle (commented):
@elisadba, thanks for the application. It is definitely an interesting concept. The architecture seems quite complex; have you considered building a system that doesn't require storing sensitive personal data? For example, I would appreciate my attorney being able to sign my transactions, but for that he doesn't need to store any identifying data anywhere. Also, any legal officer will already have some way (following national legal requirements) of storing sensitive data in place. I'm not questioning that you put a lot of thought into this, I'm mainly curious if I would I be able to use (a subset of functions of) your wallet without backend or IPFS?

gdethier:
@semuelle Thanks for your feedback and questions! We considered not storing any sensitive data given the complexity and security concern. However, 2 elements actually pushed us to not do so:

- In order to provide secure multisig and recovery services, we consider that the identity of an account's owner must be known: starting from just an e-mail and/or a phone number with multisig, up to full identity information for recovery. Otherwise, how can one prove that the person actually claiming access to an account is legitimate? Similarly, in the context of a multisig transfer, without some contact information (e-mail, phone), how could you check that the initiator is actually the owner of the account and not some malicious person which gained access to one's device and/or credentials? Therefore, identity data (which are confidential and sensitive) must be communicated to and stored by the legal officer.
- We want to ensure that the *process* of storing sensitive data is as decentralized, secure, auditable and transparent as possible. Letting the legal officers simply store the data on their hardrive may be considered as decentralized but it is not necessarily secure nor auditable or transparent. By making the storage process part of our architecture, we get closer to that goal.

semuelle:
Thanks for the reply, Gérard. I have to say I'm quite concerned about this creating honeypots, and about backing up onto IPFS in general, encrypted or not. But I do see the usefulness of trying to make the storage and backup process easier for the user. And, I suppose, making it as safe as possible is a question of configuration in the end. That's why I asked about the ability to use a subset of features. Have you confirmed that the process you are using is compatible with the legal requirements for European judicial officers' legal data storage requirements? If so, I'm happy to proceed.

BEdasc:
Hello @semuelle, thank you for your interest and feedback which I fully agree with. I have worked for more than 20 years in the judicial area. The logion project is born from a cross between this background and a strong belief in what blockchain can bring to the creation of new economic and social models, new value chain.
Our network of legal officers is bound by European legal environment. These officers would not be involved if the rules were not respected. Indeed, the sanctions in consequence would be severe. So yes, we make sure that all applicable EU regulations are respected by our processes.It is an essential element.

gdethier:
@semuelle As David mentioned, we pay a particular attention to be aligned with legal requirements. But we also share the same concern as you with regard to creating "honeypots". You'll find below a couple of technical counter-measures that we plan to implement in order to mitigate the risk as much as possible:

- The IPFS network on which the encrypted data are stored is private. Only legal officers' nodes will be able to connect to it. Encryption is essentially there because a legal officer cannot have access to the data of another legal officer.
- We plan to implement end-to-end encryption so that private keys are never processed nor stored by logion nodes.
- A rotation of private keys will be required.
- A "vault" service is planned to secure the private keys (i.e. enable the recovery of current private key in case of loss).

However, please note that the 3 last items of the list are not part of this grant proposal. They are nonetheless serious future candidates for our next steps.

BenWhiteJam (approved):
Looking forward to it.

PR #643, opened 2021-10-13
Add security.txt suggestion

semuelle:


semuelle:
> You are changing a lot more than just security.txt ;-), but looks good to me

That's mainly my markdown linter doing things (plus me making prettier headings :)). I will submit a PR with a settings.json soonish that maybe we can all use so we don't overwrite each others markdown constantly.

PR #650, opened 2021-10-15
add markdownlint settings

semuelle:
This PR adds a file that overrides VSCode user settings. I think this is generally handy as long as the majority uses vscode. But in this case, they are rules for the extension [markdownlint](https://marketplace.visualstudio.com/items?itemName=DavidAnson.vscode-markdownlint). 

This should prevent our linters from fighting.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,6 +1,6 @@
 {
     "editor.codeActionsOnSave": {
-        "source.fixAll.markdownlint": true
+        "source.fixAll.markdownlint": false
     },
     "markdownlint.config": {
         "default": true,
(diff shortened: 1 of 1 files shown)


Noc2 (approved):

.vscode/settings.json
```diff
@@ -0,0 +1,17 @@
+{
+    "editor.codeActionsOnSave": {
+        "source.fixAll.markdownlint": true
```
alxs, line 3: I'm not sure I want this as I use autosave. And some lints are better handled silencing rather than fixing them.

alxs: I meant on a per-case basis, but the sticking point is the autosave - I don't think insta-fixing every lint that shows up while editing will be a good experience

semuelle: we can try without it

```suggestion
        "source.fixAll.markdownlint": false
```

Noc2 (approved):

PR #671, opened 2021-11-02
Clarify guidelines for maintenance grants

alxs:
- MGs may only be used for small features, bug fixes and dependency upgrades.
- They must be limited in duration.


Noc2 (approved):

Noc2:
We might need to clean up the readme/move parts to other readmes at some point. It’s becoming rather long.

PR #691, opened 2021-11-12
Amend Perun Integration: Remove payment address to allow for fiat payment

matthiasgeihs:
As discussed with @semuelle , we would like to amend the proposal to allow for payment in fiat.


matthiasgeihs:
Closed for now due to some recent updates.

PR #729, opened 2021-12-05
Amendment: Jeton_Network_follow_up.md

darkfriend77:
We have been rebranding to Ajuna Network.

We would like to ask for more time for the delivery of Milestone 2, 3 & 4. (added expected date)

As we got a major change with monthly-2021-10, which allowed us to move towards a fully generated C# API, we were working on that approach and dropping the manual coded types, since we in parallel are working on Ajuna Networks Layer 2 solution, which is used for the Milestone 4 use-cases, we will need more time.  We hope we can get confirmation that we can extend the timeline for the open grant. I added the deliverables for the generator, which will be part of the grant.

If there is more information needed why we delay, then I'm happy to explain.

Best,

Cedric

### DIFF
--- a/applications/ajuna_network_follow_up.md
+++ b/applications/ajuna_network_follow_up.md
@@ -4,12 +4,12 @@
 
 > This page is also available in [Chinese (中文)](./application-template-cn.md).
 
-* **Project Name:** Jeton Network & DOTMog
-* **Team Name:** DOTMog Team
+* **Project Name:** Ajuna Network
+* **Team Name:** Ajuna Network Team
 * **Payment Address:** 1FgaH4sRHF4boR4Fw316ddKMJbFpsDXGeL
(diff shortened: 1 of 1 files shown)


Noc2 (approved):
Thanks for creating the amendment. In general it looks good to me, but it looks like you still need to sign the update CLA/terms and conditions.

darkfriend77:
signed, 👍

mmagician (approved):

alxs (approved):

semuelle:
Hey @darkfriend77. I see plenty of activity on the Ajuna repos, so I assume you are still working on this grant. Can you give us an estimate of when you are planning to submit the next milestone? It might be worth amending the contract again, or, if you don't know if you can deliver at all, we can cancel the grant for now. You would be able to reapply, of course.

darkfriend77:
@semuelle Yes, will deliver at the start of July, after the Decoded.

PR #739, opened 2021-12-13
Candle Auction Milestone-3: Workshop to substrate.dev

agryaznov:
### Project Abstract

This is a proposal to continue the project with **Milestone-3:** adding a Workshop to substrate.dev

### DIFF
--- a/applications/candle_auction_ink.md
+++ b/applications/candle_auction_ink.md
@@ -114,7 +114,7 @@ which contains both sources and documentation.
 
 ### Milestone 2 - Random close
 
-* **Estimated Duration:** 1 weeks
+* **Estimated Duration:** 1 week
 * **Costs:** 2500 DAI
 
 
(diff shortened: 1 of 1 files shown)


Noc2 (changes requested):
Thanks for the application. In general, we usually don’t support grants that focus only on education/tutorials, but since this is an update to your previous grant application. I would be willing to go ahead with it. Could you maybe add a license to your deliveries. Also since we can not ensure that your delivery is actually merged into https://docs.substrate.io/

agryaznov:
Hi David. The reason I haven't added a license to Milestone-3 is that the [substrate-docs repo doesn't specify the license itself](https://github.com/substrate-developer-hub/substrate-docs#license). So I'm not sure if *Apache 2.0* would be ok here, not to contradict with the target site (substrate.dev) license.

Please, kindly advise on it. If Apache 2.0 wouldn't be an obstacle there, I'll add just it. Thx.

semuelle:
I think [this](https://docs.github.com/en/github/site-policy/github-terms-of-service#6-contributions-under-repository-license) applies here, i.e. it assumes the substrate-docs license when you submit a PR.

Noc2 (changes requested):
Good point, I wasn’t aware of this. I recommend to add a delivery that says “License: either merged into substrate-docs or published under apache 2 as part of a separate repo” or something similar.

agryaznov:
Please find added license terms.

semuelle (approved):

Noc2 (approved):
Thanks for the update. Looks good

PR #778, opened 2022-01-12
Apply for Grant, Plus: Polkadot js Plus extension

Nick-1979:
### Project Abstract

Plus extension will be a user-friendly wallet to interact with the Polkadot/Substrate based blockchains through a browser. It allows users to access their Polkadot account(s), which can also be used to interact with decentralized apps. It is actually polkadot js extention, plus some new features. We don't want to rebuild the wheel, we just want to perfect it, standing on the shoulders of giants like "polkadot js extension" and make it user-friendlier for users.

The reason why we have decided to work on this project is every day user comments on social medias that complain "why the extension does not show the balance?", "It is too complicated for new/average users to work with", "It is too abstract", " why should we go to another web site to work with the extension", " how could contribute to the crowdloans via extension", and so on. That's why we've decided to work on Plus extension to address such concenrs/requests of the polkadot community. The new functionalities, that will be added to the original polkadot js extension are:

- View balances
- Transfer funds
- Easy staking (auto stake/unstake/redeem funds and edit nominated validator list)
- Contribute in crowdloans
- View transaction history
- View an address as a QR code
- Governance (Referendums/Proposals voting, Treasury Tips, councils/motions )

- **Level 1**:  Up to $10,000, 2 approvals


Noc2 (changes requested):
Thanks for the application and the nice call. I just noticed that you updated the deliveries  0a-0d of the template. Usually these deliveries “Documentation, Testing Guide, Docker” are mandatory for all grants. I think Docker isn’t necessarily required in your case, but it would be nice if you could add Documentation and a testing guide, see https://github.com/w3f/Grants-Program/blob/master/applications/application-template.md#milestone-1-example--implement-substrate-modules 
Specifically: “Deliverables 0a-0d are mandatory for all milestones, and deliverable 0e at least for the last one. If you do not intend to deliver one of these, please state a reason in its specification (e.g. Milestone X is research oriented and as such there is no code to test).”

Nick-1979:
The deliveries are updated, Thank you.

Noc2 (approved):
Thanks! I’m happy to go ahead with it and I will share the application with the rest of the team.

laboon (approved):

PR #286, opened 2021-02-28
stable-asset.md

frankyin-factual:
# Grant Application Checklist

- [x] The [application-template.md](https://github.com/w3f/Open-Grants-Program/blob/master/applications/application-template.md) has been copied, renamed ( "project_name.md") and updated.
- [x] A BTC or Ethereum (DAI) address for the payment of the milestones is provided inside the application.  
- [x] The software of the project will be released under the Apache license version 2.0 as specified in the [terms and conditions](https://gist.github.com/Noc2/75bc58e8ce9b5d419ff883b0cf2b8c19).
- [x] The total funding amount of the project is below USD $30k for initial grant applications and $100k for follow-up grants.
- [x] The initial PR contains only one commit (squash if needed before submitting your PR).
- [x] The grant will only be announced once we've successfully [delivered](https://github.com/w3f/Grant-Milestone-Delivery) the first milestone.

PR #804, opened 2022-02-03
Update stable-asset.md

shengda:
#286

- [x] **Level 2**:  Up to $50,000, 3 approvals


shengda:
The reason for the amendment is, during our deployment of milestone 1 and 2 to Mandala, we found that support of yield asset and amplification update are extremely import for Polkadot ecosystem assets.

We spent more than three weeks to implement, test and deploy our solutions in Mandala. It works well and we are planning to deploy to Karura early next week. We believe it’s sufficient as milestone 3 deliverable.

As for stable savings, we are going to apply it as a separate grant since its work scope is bigger than a single milestone, especially with XCM integrated.

alxs:
@shengda sounds good, thank you for the update. Could you provide a few more details on those two items? What do you mean exactly by yield assets, and what are amplification parameters? Ideally, please update the summary on lines 51-53.

shengda:
I have removed Stable Saving from the project details, and provided more details on the milestone 3 deliverables.

alxs (approved):
Thanks, looks good to me! I'll go ahead and share it with the rest of the team.

Noc2 (approved):

takahser:
@shengda Thanks for the amendment proposal.

Quoting from the application document:

> Comprehensive tests that cover yield asset and _amplification parameter_ change

A) I'm still confused about the meaning of _amplification parameter_. Could you elaborate on that?

> Since all assets in the baskets are of the same value peg, their prices should fluctuate about the peg prices with low variation expected so that the overall collateral ratio should be slightly over 100%

B) I was just wondering why you expect it to be >100% and not around 100% since the fluctuations obviously can go both directions. Note that this is not relevant for the acceptance of the amendment since your application was already accepted, I'm just being curious. Apologies in case this was already discussed. :)

PR #848, opened 2022-03-01
Change the scope of work of the 5th milestone in OpenBrush

xgreenx:
We want to change the scope of work of the 5th milestone in OpenBrush. 
We've changed point 2: "Add support of default implementation in trait definition on ink! level" -> "Re-work the storage of contracts"

We consider that [reworking of the contract's storage](https://github.com/paritytech/ink/issues/1134) should be done before default implementation in traits and it is more important. We participated in the discussion and we have a clear vision of how to implement that.

That change requires the same human resources as default implementation in traits so the cost of the milestone is the same.
We moved default implementation in traits to the third Grant.

### DIFF
--- a/applications/openbrush-follow-up.md
+++ b/applications/openbrush-follow-up.md
@@ -242,14 +242,14 @@ PSP - https://github.com/w3f/PSPs/pull/25
 * **FTE:**  2
 * **Costs:** 15,000 USD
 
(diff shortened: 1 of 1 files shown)


Noc2 (approved):
Sound good and thanks for the update

takahser (approved):
Sounds good to me as well. Thanks guys! 👍

PR #796, opened 2022-01-27
Create Tao-Framework.md

taocodao:
Grant Application Form for Tao-Framework project

### Project Abstract

> Please provide a brief description of your project here summarising key points (1-2 paragraphs).
>
> If your application is a follow-up to a previous grant, please mention which one in the first line of the abstract and include a link to previous pull requests if applicable.

### For which grant [level](https://github.com/w3f/Grants-Program#level_slider-levels) are you applying? 
- [ ] **Level 1**:  Up to $10,000, 2 approvals
- [x] **Level 2**:  Up to $50,000, 3 approvals
- [ ] **Level 3**:  Unlimited, 5 approvals (for > $100k Web3 Foundation Council approval)

### Application Checklist

- [x] The [application template](https://github.com/w3f/Grants-Program/blob/master/applications/application-template.md) has been copied, renamed ( `project_name.md`) and updated.
- [x] A BTC or Ethereum (DAI/USDT) address for the payment of the milestones is provided inside the application.
- [x] I have read and acknowledged the [terms and conditions](https://github.com/w3f/Grants-Program/blob/master/docs/T&Cs.md).
- [x] The software delivered for this grant will be released under an open-source license specified in the application.
- [x] The initial PR contains only one commit (squash and force-push if needed).
- [x] The grant will only be announced once the first milestone [has been accepted](https://github.com/w3f/Grant-Milestone-Delivery#process).

PR #876, opened 2022-03-22
Tao: Asynchronous Payment Channels to Transfer Value in the Web3 World 

taocodao:
### Project Abstract

Tao is designed to overcome blockchain technology's ultimate challenges: scalability and privacy. The public blockchain, such as Polkadot, has the limited transaction throughput associated with the underlying consensus protocol, which demands that every system participant verifies and stores a replica of the entire history of transactions. Also, blockchain applications require low overhead to be useable, requiring their transaction costs remain low. A bottleneck for this has been on-chain payments, which incur a fee for every transaction, especially when we need many micro-transactions to be sent. Off-chain solutions seem promising since they only use the blockchain as a fail-safe mechanism and move the transaction load offline, where the bottleneck is network speed. Any two parties that interact can set up a joint account on the blockchain, a "channel" through which the two parties can transact off-chain, sending money back and forth. Two parties rely on the blockchain as a fail-safe mechanism in disputes. Current proposals, however, require synchrony assumptions to preserve the channel safety, leaking to an adversary the exact amount of time needed to control the network for a successful attack.

Therefore, we introduce "Tao," an innovative payment channel solution that remains secure under network asynchrony and concurrently provides correct incentives. The core idea is to incorporate the conflict resolution process within the channel by introducing a rational committee of external parties. Hence, if a party wants to close a channel unilaterally, it can only get the committee's approval for the last valid state. Furthermore, to provide sub-second latency, It uses consistent broadcasts to announce updates and close the channel. This lightweight abstraction is powerful enough to preserve safety and liveness to any rational parties.

Tao is a plugin-based system to provide a low-cost micropayment service to the DAPPs running on the Polkadot network. A plugin is developed using web3.js that allows DAPPs to interact with a Moonbeam node through Ethereum-style JSON RPC invocations. By running the smart contracts deployed on the Moonbeam nodes, an off-chain payment channel can be open between the buyer and seller. Thus Dapps can provide a "pay as it goes" service.

- [X] **Level 2**:  Up to $50,000, 3 approvals

### Application Checklist

- [x] The [application template](https://github.com/w3f/Grants-Program/blob/master/applications/application-template.md) has been copied, renamed ( `project_name.MD) and updated.


semuelle:
Hi @taocodao, thank you for your application. 

Do I see correctly that you are applying for three separate grants? I'm afraid we only allow one grant at a time per team. You could either apply for one now and follow-up grants after completion, or apply for one large grant together. For the former, you would have to remove two of the files from this PR. For the latter, you also would have to remove two files, but you could rewrite the remaining one to include the contents, milestones and funds of the other two. Note that all three grants together (>100,000 USD) would be a Level 3 grant and require the approval of five committee members and the W3F council.

Let me know if you have any further questions!

taocodao:
Thx very much for the feedback and suggestion from Sebastian. We really appreciate it. We apologize for the confusion since we are not very familiar with Github and the pull request process. We only apply for one grant, and the code is Tao-Blockchain.md. We have already closed the other two requests, and have removed the other two files accordingly now. Please confirm. thx

semuelle:
Thanks for the update, @taocodao. Could you move the file to the `applications` subfolder? We will continue reviewing the application in the next few days.

taocodao:
We have done intensive research on blockchain technology for the last three years. We think a good blockchain network design should be decentralized, scalable, interoperative, and self-sustaining ( self-funded). These are the principles we have followed when we design the Tao Blockchain. So please give us some feedback and suggestions to help us improve. Thx very much.

semuelle:
Dear @taocodao, sorry for the long wait. Your application was not forgotten or ignored, we are just experiencing a high load of applications and deliveries. I have read your application(s), and I have a couple of comments/questions:

- Why did you close your previous pull request (https://github.com/w3f/Grants-Program/pull/796)? Has the application changed fundamentally since then?
- We usually expect someone from the team to have provable experience writing and maintaining open source software. There is not a lot of activity from your Github account and no other team members mentioned, and your website returns 404. We often see teams struggling that don't already have the necessary manpower or skills. Due to the high demand of developers, we also have no way of helping you there.
- If your nodes are based on Moonbeam, are the smart contracts written in Solidity? We don't offer funding for Solidity smart contracts as this is not our expertise. Have you had any luck with your Moonbeam application?
- The specification of deliverables should contain enough information to verify their completeless after delivery. Many deliverables are currently too vague. For example, what does it mean to "integrate all the development components"? What exactly is being done here? What is included when you build "a prototype for a specific use case"?

Noc2 (changes requested):
Thanks for your new application. However, the deliveries of your milestone need to be very clearly defined. For example, in which programming language will they be implemented, what exactly are VPOS; Token Signers; Communication Channels. What do you mean with etc. What smart contract are you going to implement, regarding “We will develop a public smart contract”. What is TNPOS and how is it different from a normal proof of stake solution? How is the standard token model different from existing standards like ERC20, etc.? In general, my recommendation would be to focus on one problem with this grant application and initially only apply for this. For example, if you want to create a new kind of token standard, maybe focus on this and implement it as an ink smart contract or pallet.

taocodao:
@Noc2 @semuelle Thx very much for your feedback and suggestion. We have re-visit our plan multiple times, and add a lot of detail to the implementation plan. Our goal is simple: we try to leverage blockchain technology and bring it to our daily life and make a real impact. Please continue to help us by giving us more advice. Thx

taocodao:
Thx very much for the feedback. Yes, our application targets the Polkadot from day one, we only use Ethereum as an example for the public blockchain network. We have updated the application according to your suggestion, please review, thx

Noc2 (changes requested):
Thanks again for the update. Something to be aware of is that we usually don’t support the development of solidity smart contracts via our grants program. On the one hand are we no experts when it comes to solidity on the other hand is it hard for us to judge what is new and what does already exist. Additionally to clarify this, could you also add the programming languages to the different deliverables.

taocodao:
Thx very much for the feedback. We will use the standard ERC-777 implementation in Openzeplien for the token implementation, other smart contracts will be implemented in ink! or developed as the pallet. Thx.

Noc2 (changes requested):
Thanks! Could you add this information to the specification of each deliverable?

taocodao:
Thanks very much for your suggestion. As you suggested, we have done a significant update on our proposal to focus on one problem at a time. For this grant application, we only keep the offline payment section. Please review. thx

Noc2 (changes requested):
Thanks for the update. I will share the application with the rest of the team. But I have a few additional questions: Why do you want to use the moonbeam tech stack, when you plan to develop ink smart contracts? In general it’s helpful to specify the programming language as part of the specification.

taocodao:
@Noc2 Thanks so much for sharing. We are using the MoonBeam tech stack because MoonBeam itself is a public blockchain and provides a platform to run smart contracts. Our solution will be implemented as the smart contract follows the algorithm specified in the doc on the channel open, update, and close. We have already implemented part of the protocol in Solidary. I will update that in the Implementation status section. If those smart contracts run well on the MoonBeam dev nodes, we will keep them in solitary; otherwise, we will start implementing them in Ink!. Let us know what you think. Thx again for all your help! Cheers!

Noc2 (changes requested):
Sorry for the late reply here. I thought I already did reply before, but apparently I forgot to click on send. You probably mean Solidity instead of Solidary, correct? However, we don’t support the development of solidity projects via our grants program. One reason is that we are no experts when it comes to solidity. Another reason is that we don’t know what already exists at this stage and what needs to be developed, etc. You could either potentially apply for a moonbeam grant: https://moonbeam.foundation/grants/ or maybe you are interested in an ink! Smart contract implementation instead.

taocodao:
@Noc2 Thanks very much for the suggestion. Yes, we are very interested in an ink! Smart contract implementation. We have all the protocols developed in pseudo-code. So we will convert them into Smart Contracts using ink! We will update the application to reflect that as well. Thx

Noc2 (changes requested):
Thanks for the update. In this case also feel free to remove “MoonBeam dev node” from the application, since moonbeam doesn’t support ink! Smart contracts as far as I know. And if I understand it correctly, Consistent Broadcast 5d is basically your backend for the state channels, correct? In which programming language will you develop this (feel free to add this to the application)? Regarding your second milestone, do you already know which ​​ProtoType you want to develop. Otherwise it might make sense to initially only focus on the first milestone and apply for a follow-up grant later.

taocodao:
@Noc2 Thx very much for the advice, we have updated the document accordingly. Please review thx

Noc2 (changes requested):
Thanks for the update. However, why did you increase the price of the first milestone? And it’s still not fully clear to me what “consistent Broadcasting” is? Is this the backend for the state channels? In which programming language are you going to implement this?

taocodao:
@Noc2 Thx very much for the feedback. We have a serious discussion among ourselves. We decided to put back milestone 2, since we have a good use case and have added that into the development status section. Please review. Regrading  "consistent broadcasting," it is a lightweight consensus protocol used to build the committee of validators and will be implemented using the smart contract developed using ink!. You are right. It is used to build the backend of the state channel. Please let us know if you have further questions. Thx

Noc2 (changes requested):
Thanks again for the reply. However, do you also have parts that don’t require a consensus for the payment channels and a simple centralised off-chain backend? Otherwise it sounds a little bit like you want to scale by adding an additional blockchain to an existing blockchain (including decreasing the overall security). Also you might want to add additional (technical) information about the prototype to the second milestone.

taocodao:
@Noc2 Correction, the consistent broadcasting is only used during the channel update, not in the close and open. thx

Noc2 (changes requested):
Thanks! But for the payment channels, you still need some off-chain database or something similar, don’t you? Or are you fully relying on peer to peer connection (which might be slow)?

semuelle (changes requested):
- What is the purpose of the `private network` (Deliverable 2b)? Who are its participants and what consensus algorithm does it use? Why is it private?
- Can you explain in simple terms what the relationship between the Moonbeam nodes and the Substrate network/ink! contracts is? Are they parallel implementations? It doesn't make sense to me to have both, unless I misunderstood something fundamental about your architecture. Perhaps an architecture or flow diagram would help.
- As I mentioned [earlier](https://github.com/w3f/Grants-Program/pull/876#issuecomment-1085155970), we usually expect grantees to have provable experience with Substrate or similar technologies. However, I don't see any activity on your Github account nor any other proof of anyone being (active) on your team. Can you elaborate what you mean by "extensive expertise on multiple levels"?

taocodao:
@Noc2 @semuelle We have updated our document to illustrate the detailed process in a fun way, please refer to the following link: https://docs.google.com/document/d/15pOrIl6cI2gj2cJWrOHOFEd-m0x49ZjrHbXXnhZfYGw/edit?usp=sharing. Please let us know if you have any further questions, thx

semuelle:
Thanks for the update, @taocodao, and sorry for the delayed reply.

> The participants are two parties to the transactions and the committee of validators. 

- "Private" network, to me, is a network where the public cannot simply join. Who are the validators? How does one become a validator on the private network?
- How do I know that I can trust the validators, and which validators?
- What happens when I open a channel with a selected set of validators, and some or all of those validators go offline?

taocodao:
@semuelle Thx very much for the questions. See the answers below:

Question: "Private" network, to me, is a network that the public cannot simply join. Who are the validators? How does one become a validator on the private network?
Answer: The core idea is our design is to provide proactive security to the channel instead of reactive dispute resolution. A group of validators and the channel parties form the "private network", If there is a dispute, the Validators make sure the correct state is the only one available for submission on-chain, regardless of the amount of time it takes to make this final state visible. They are running a lightweight consistent broadcast protocol to preserve both safety and liveness. 

Question: How do I know that I can trust the validators, and which validators?
Answer: The Validators commit their identities on the blockchain during the funding transaction of the channel open. Both parties of a channel have to agree on a committee of Validators before opening the channel. 

Question: What happens when I open a channel with a selected set of validators, and some or all of those validators go offline?
Answer: Regarding the committee of the validators, we assume that there are at most f out of n = 3f + 1 Byzantine Validators( malicious or offline) and we define a threshold t = 2f + 1 to achieve the liveness and safety properties.

if you have any further questions don't hesitate to let us know. Thx again for all your help!

semuelle:
> The core idea is our design is to provide proactive security to the channel instead of reactive dispute resolution. A group of validators and the channel parties form the "private network"

That doesn't answer the question who the validators are. Can anyone become a validator? Can you expand on what it means for a "group of validators and the channel parties [to] form the private network"? What exactly happens in the background? It sounds as if you are creating an ad-hoc blockchain comprised of a handful of validators and two users.

> The Validators commit their identities on the blockchain during the funding transaction of the channel open. Both parties of a channel have to agree on a committee of Validators before opening the channel.

How does that work? Will there be a catalogue of validators hosted somewhere, that I can look at, with their identity information and statistics as validator? Possibly on a website you host?

taocodao:
@semuelle @Noc2 : Please see the following as the channel process detail:

1. Implemented the on-chain operations in the form of a smart contract in ink!
2. The contract allows the parties to specify the number n of Validators they desire to involve and their identities.
3. The signature scheme makes off-chain signatures compatible with on-chain accounts, and as such, signatures made off-chain can be verified using public keys available on-chain.
4. Once the contract is deployed on the Ethereum network, Alice funds it first.
5. Subsequently, once Alice's funding transaction is finalized, Bob funds it.
6. Once Bob's funding transaction is finalized, the collateral can be calculated, so the Validators can fund it in any order simultaneously.
7. When all Validators have funded the contract, any of the two parties can open the channel.
8. At any time before opening the channel, any party or Validator can withdraw their money. At this point, the channel is canceled and can no longer be opened but allows the rest of the parties to withdraw in any order.
9. Once the channel is open, the parties can continue exchanging states off-chain.
10. If multiple channels are used, then the cost of smart contract deployment can be amortized by abstracting the common functionality into a Solidity library. However, the opening and closing costs are recurrent.
11. When the parties wish to close the channel optimistically, Alice initially submits a transaction to the smart contract requesting the channel to close. This request contains Alice's claimed closing state (namely, Alice's value at closing time, as Bob's value at closing time can be deduced from this.
12. Once Alice's transaction is confirmed, if Bob is in agreement, he submits a transaction to the smart contract to signal his agreement.
13. The smart contract then returns Alice's and Bob's values and Validators' collateral. Check that the sum of Alice's value and Bob's value at the closing state does not exceed the sum of their values at their initial state so that sufficient funds remain to return the Validators' collateral.
14. If Bob does not agree with Alice’s claim, the channel becomes unusable and must be closed pessimistically (Alice can no longer make an optimistic claim on a different state).
15. The optimistic close operation measures the cumulative gas cost of the two transactions from both parties. The cost is minimal and should be the normal path since parties need not pay closing fees.
15. Each Validator then submits a transaction to the smart contract containing the sequence number they have seen last, together with Alice’s and Bob’s off-chain signature on it. These can be submitted in any order.
17. The signatures of Alice and Bob on the sequence number are verified on-chain; this incurs the majority of the gas cost for the pessimistic close. The party who wishes to close the channel monitors the chain for such claims and remembers any fraudulent ones.
18. As soon as t (honest or adversarial) claims have been recorded, either Alice or Bob can send a transaction to the smart contract to close the channel.
19. The transaction is accompanied by the fraud proofs the closing party could assemble, namely the latest announcement for each Validator who claimed an earlier sequence number.
20. These announcements contain the signature of the Validator on the plaintext, which consists of the smart contract address and the sequence number.
21. These announcements contain the signature of the Validator on the plaintext, which consists of the smart contract address and the sequence number.
22. The pessimistic close operation was measured when no fraud proofs are provided and included the transaction of each of the t Validators and the final transaction by one of the parties. Additionally, it was assumed that, while the counterparty is unresponsive or malicious, the Validators were responsive and all submitted the same sequence number (hence limiting the need for multiple signature validations).

taocodao:
@semuelle @Noc2 Please review and let us know if you have further questions. Thx

semuelle:
Thanks for the update, and again, sorry for the late reply. 

> a lightweight, consistent broadcast protocol

I'm afraid the description of your project is still very opaque to me. I suppose, what I would need is an architecture diagram, that shows what functionality is implemented where. When you use the term Validator, it sounds like a parachain validator, and that is one of your deliverables, but that is not consistent with how you are describing the process. How do you enable individual nodes only to prevent illegitimate and approve legitimate smart contract executions? And how is the validator selection process implemented?

I am going to share this again with my colleagues, but I'm afraid if no one voices their support for it in the next ten days, we will have to close this PR.

taocodao:
@semuelle Thx very much for the question. I think the name "validator" is confusing; we should have called it "Warden" After opening the channel on the blockchain, the channel can only be closed either by a transaction published on the blockchain and signed by both parties or by a transaction signed by one of the parties and a threshold (t) of honest Wardens. Thus, the committee acts as
power of attorney for the parties of the channel. Furthermore, it employs correct incentives for the t rational Wardens to follow the protocol, hence it can withstand t = 2f + 1 rational and f Byzantine Wardens, while the wealthiest channel party is assumed rational and the other byzantine.
A naïve solution would then instruct the committee to run asynchronous consensus on every new update, which would cost O(n4) per transaction, a rather significant overhead for the critical path of our protocol5. Instead, in Tao, the consensus is not necessary for update transactions, as we only provide guarantees to rational parties (if both parties misbehave, one of them might lose its funds). As a result, every time a new update state occurs in the channel (i.e., a transaction), the parties run a consistent broadcast protocol (cost of O(n)) with the committee. Specifically, a party announces to each Warden that a state update has occurred. This announcement is a monotonically increasing sequence number to guarantee that the new state is the freshest state, signed by both channel parties to signal that they agree. If the consistent broadcast protocol succeeds (t Wardens acknowledge reception), this can prove to both parties that the state update is safe. After this procedure terminates correctly, both parties proceed to execute the off-chain state. 
At the end of the life cycle of a channel, a dispute might occur, leading to the unilateral closing of the channel. Even in this case, we can still guarantee the security and liveness of the closing with consistent broadcast. The core idea is that if 2f + 1 Wardens accepted the last sequence number before receiving the closing request (hence the counterparty has committed), then at least one honest Warden will not accept the closing at the old sequence number. Instead, the Warden will reply to the party that it can only close at the state represented by the last sequence number. As a result, we define a successful closing as the maximum of all proposed states, which guarantees safety. Although counterintuitive, this closing process is safe because the transactions are already totally ordered and agreed to by the parties of the channel; thus, the committee acts as a shared memory of the last sequence number.

PR #939, opened 2022-05-11
add option for private discussion

semuelle:
Let me know if you think there is a better way of doing this.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -19,6 +19,7 @@
 - [ ] The software delivered for this grant will be released under an open-source license specified in the application.
 - [ ] The initial PR contains only one commit (squash and force-push if needed).
 - [ ] The grant will only be announced once the first milestone [has been accepted](https://github.com/w3f/Grant-Milestone-Delivery#process).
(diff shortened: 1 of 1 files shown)


Noc2 (approved):
Looks good to me.

Noc2:
I also shared it with hakan.

laboon:
Looks good to me

hakan-w3f:
Maybe the form could ask for a set of usernames so that an entire team could communicate with us? Or would each team set up a group alias in Matrix? I'm not knowledgeable in Matrix, so not sure how an alias would work for a team etc. Please advise.

takahser (approved):
lgtm

hakan-w3f (approved):
All clear now, after having read the explanation in the welcome note. Nice!

semuelle:
> Maybe the form could ask for a set of usernames so that an entire team could communicate with us? Or would each team set up a group alias in Matrix? I'm not knowledgeable in Matrix, so not sure how an alias would work for a team etc. Please advise.

The space is set up in such a way that external people can invite more people. That way, we only need to know the first one, who can then invite the rest of the team. It could even be a throwaway for those who don't want their Element handle to be public.

PR #986, opened 2022-06-10
Grant application - Review required

hoho-yin:
Hello we have revised our application according to the golden feedback 8 days ago, but no response since then. Could you please kindly have a look and make the final judgment? Thanks a lot!


Noc2 (changes requested):
Thanks for chasing us. But as pointed out multiple times now, please don’t open new PRs for this. Simply ping us on the old PR. This doesn’t help with your grant application.

PR #990, opened 2022-06-15
Remove polkadot-java from tech stack

alxs:
Because, in another team's [words](https://github.com/strategyobjectadmin/Grants-Program/blob/master/applications/substrate_client_java.md#relevance-to-substrate--polkadot), it is "totally outdated, thus absolutely unusable."


Noc2 (approved):
Happy to remove it, because we have other libraries. But maybe we should consider pointing out projects that are outdated and we would be willing to support maintenance or getting it updated again.

alxs:
That would make a lot of sense. Just in this case, as you say we have other options and also I think last time we discussed it we didn't think it was worth it to pick up work on this again.

PR #1026, opened 2022-07-05
Add status badge action

lucasvanmol:
Adds status badges to links in the polkadot stack using https://github.com/lucasvanmol/status-badges

While testing on my fork it looks like I was rate limited, hence all the question marks because I was getting `403: Forbidden`. The merge to this repo should mean a seperate `GITHUB_TOKEN` will be used for authentication to the github API, so all the badges should get updated normally once merged. 

Currently, the github action runs and updates all the bages:
- every day at midnight
- whenever `docs/polkadot_stack.md` gets changed on push
- when run manually through the actions tab on github.

I can try to implement a fix for the rate limiting issue, so that it doesn't overwrite existing emojis when it does happen. In the mean time, rate limiting shouldn't cause too many problems, unless `docs/polkadot_stack.md` is getting committed to very frequently. The workflow can also be changed so that it doesn't get updated when the stack does, and only at midnight or when run manually, if that's preferable.

Additionally, there are some problems with links that have redirects and are returning [301](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/301), I assume due to some teams that have e.g. changed their github name. It causes the action to not find them and set their status as "❔" even though the repo is there. I can also try to fix this in the action, but in the mean time we should just be able to update the links to the new locations and it should work fine again.


Noc2 (approved):
I’m happy to give it a try. Great work!

PR #872, opened 2022-03-17
Create tdot.md

shengda:
### Project Abstract

This is a follow-up of the [stable asset grant](https://github.com/w3f/Grants-Program/pull/286).

tDOT is a DOT-pegged derivative built on top of the stable asset protocol. It aims at generating unified DOT liquidity across Polkadot parachains and applications.

### For which grant [level](https://github.com/w3f/Grants-Program#level_slider-levels) are you applying? 
- [ ] **Level 1**:  Up to $10,000, 2 approvals
- [x] **Level 2**:  Up to $50,000, 3 approvals
- [ ] **Level 3**:  Unlimited, 5 approvals (for > $100k Web3 Foundation Council approval)

### Application Checklist

- [x] The [application template](https://github.com/w3f/Grants-Program/blob/master/applications/application-template.md) has been copied, renamed ( `project_name.md`) and updated.
- [x] A BTC or Ethereum (DAI/USDT) address for the payment of the milestones is provided inside the application.
- [x] I have read and acknowledged the [terms and conditions](https://github.com/w3f/Grants-Program/blob/master/docs/T&Cs.md).
- [x] The software delivered for this grant will be released under an open-source license specified in the application.
- [x] The initial PR contains only one commit (squash and force-push if needed).
- [x] The grant will only be announced once the first milestone [has been accepted](https://github.com/w3f/Grant-Milestone-Delivery#process).

PR #1079, opened 2022-07-19
Update tdot.md

shengda:
https://github.com/w3f/Grants-Program/pull/872

We propose amendment to tDOT grant because we enhance our original architecture design.
* During the implementation, we find it more robust and flexible to introduce the concept of local LP which is generated by individual stable asset pool and aggregate LP which is generated by local LP. tDOT is an aggregate LP. Therefore, we add a milestone for implement stable asset XCM pallet to manages both LPs;
* Minting and redeeming tDOT on host chains requires changes to the stable asset pallet due to change in milestone 1. Therefore, we expand the scope of milestone 2 and 3;
* We remove cross-chain swap from the grant since a cross-chain swap can be seen as a combination of cross-chain mint + redeem. We are going to implement such swap in DApp instead of on pallet;
* We remove cross-chain fee and yield handling from the grant since we have implemented an off-chain fee and yield handing solution. Since this solution is offchain and we have run it successfully on Kusama and Polkadot, we decide to remove them from the grant.


cruikshankss:
Good day @shengda, thank you for your application. I will look into it as soon as possible.

applications/tdot.md
```diff
@@ -101,68 +101,71 @@ NUTS Finance is a blockchain development DAO. Our team is composed of experience
 - **Full-Time Equivalent (FTE):**  2
 - **Total Costs:** 16,000 DAI
```
semuelle, line 102: If I'm counting correctly, this should now be

```suggestion
- **Total Costs:** 14,000 DAI
```

semuelle (changes requested):
Thanks for the amendment, @shengda. It sounds as if you are now planning to implement everything in one pallet, instead of several. Is that correct?

Noc2 (changes requested):
I’m happy to approve it, once the total costs are updated.

shengda:
> 

We will implement it in two pallets.
* Stable asset pallet is the upgraded version from our previous grant. It will be deployed on each parachain that supports stable asset and mints local LP;
* Stable asset XCM pallet manages all cross-chain minting and redemption. It will be deployed on the host chain of the aggregate LP.

semuelle (approved):
Thanks for the update. Happy to proceed.

Noc2 (approved):

alxs (approved):

PR #1103, opened 2022-08-03
Simplify pull request template

alxs:
- FAQ shouldn't be mandatory
- We have the CLA bot for the T&C's
- The announcement guidelines are shared again in the approval comment, and there is already a checkmark for not announcing before the first milestone
- Added default homeserver since many people don't know what it is
- We already ask where they heard about us in the template, if we do want to have it here we can remove it there instead

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,32 +1,21 @@
-### Project Abstract
+#### Project Abstract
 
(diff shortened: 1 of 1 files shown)


Noc2 (approved):
Sounds good. I'm always a fan of removing text that no one is reading anyway.

PR #1133, opened 2022-08-21
Update GreenLemon.md

wuyahuang:
The milestone 1 deliverable mentions ApprovalForAll and set_approval_for_all, however, these functions are not required because each token owner is a different, one-time encrypted address. So these two functions need to be removed.

We added 2 additional features to this milestone, which are both not mentioned by the deliverable. First, add function token_uri for users to query the link of NFT metadata. Secondly, we add an additional param into the message before hashing, which is the token nonce. Account nonce is unfit due to each token owner being a unique and one-time encrypted address. When an NFT is operated once, its corresponding token nonce is automatically added by 1. We think that token nonce can prevent replay attacks for signatures already sent to the blockchain.


Noc2 (approved):
Thanks for the update. I’m happy to approve this and share it with the rest of the team.

takahser (approved):

alxs (approved):

laboon (approved):

PR #1165, opened 2022-09-06
TREX - Timed Release Encryption Xing-chains

yangfh2004:
#### Project Abstract
TREX is a project to develop a protocol and network to support permissionless timed-release encryption in Web3. It is a decentralized network that consists of validators, archivers, off-chain workers in trusted execution environment (TEE), and other essential substrate nodes. This project is developed based on the Substrate framework as a parachain of the Polkadot ecosystem. We aim at institutional and business clients who need blockchain technology to enhance their data security and build [decentralized reputations](https://newsletter.thedefiant.io/p/-decentralized-reputation-is-about-a8b) on Web3.

#### Application Checklist

- [x] I have read the [application guidelines](https://github.com/w3f/Grants-Program/blob/master/docs/grant_guidelines_per_category.md).
- [x] A BTC, Ethereum (USDT/USDC/DAI) or Polkadot/Kusama (aUSD) address for the payment of the milestones is provided inside the application.
- [x] The grant will only be announced once the first milestone [has been accepted](https://github.com/w3f/Grant-Milestone-Delivery#process) (see the [announcement guidelines](https://github.com/w3f/Grants-Program/blob/master/docs/announcement-guidelines.md)).
- [x] I prefer the discussion of this application to take place in a private Element/Matrix channel. My username is: `@yangfh2004:matrix.org` (change the homeserver if you use a different one)

### DIFF
--- a/applications/TREX_Network.md
+++ b/applications/TREX_Network.md
@@ -3,105 +3,94 @@
 - **Project Name:** TREX - Timed Release Encryption Xing chains
 - **Team Name:** [NexToken Technology](https://github.com/NexTokenTech)
 - **Payment Address:** 0x10286c9d17a1aF1D73E19762A5132d604a076585 (USDT)
-- **[Level](https://github.com/w3f/Grants-Program/tree/master#level_slider-levels):** 2
+- **[Level](https://github.com/w3f/Grants-Program/tree/master#level_slider-levels):** 1
 
 ## Project Overview :page_facing_up:
 
-TREX is a project for the development of a protocol and network to support trustless and permissionless timed-release encryption in Web3. It is a decentralized network that consists of miners, archivers and basic substrate nodes. This project is developed based on the Substrate framework as a parachain of the Polkadot ecosystem. We aim at institution and business clients who need blockchain technology to enhance their data security and build up [decentralized reputations](https://newsletter.thedefiant.io/p/-decentralized-reputation-is-about-a8b) on Web3.
+TREX is a project to develop a protocol and network to support permissionless timed-release encryption in Web3. It is a decentralized network that consists of validators, archivers, off-chain workers in trusted execution environment (TEE), and other essential substrate nodes. This project is developed based on the Substrate framework as a parachain of the Polkadot ecosystem. We aim at institutional and business clients who need blockchain technology to enhance their data security and build [decentralized reputations](https://newsletter.thedefiant.io/p/-decentralized-reputation-is-about-a8b) on Web3.
 
 ### Overview
 
-TREX is the world’s first blockchain platform that enables decentralized timed-release encryption (dTRE) to unlock information 
-(as serialized and structured data) and actions (as serialized procedures to be called in distributed remote nodes) 
-for future events. With TREX, the data and procedure were encrypted and distributed in a decentralized blockchain network. 
-The data or procedure would be only revealed or executed after certain counts of clock ticks. 
+TREX is the world’s first blockchain platform that enables decentralized timed-release encryption (dTRE) to unlock information (as serialized and structured data) and actions (as serialized procedures to be called in distributed remote nodes) for future events. The data and procedures were encrypted and distributed in a decentralized blockchain network. The data or procedure would be only revealed or executed after specific counts of clock ticks. 
 Based on a monetary incentive mechanism, all nodes in the TREX network will collaborate and automatically adjust to ensure 
 the accurate release of data and procedures as a consensus-based system.
 
-Our consensus is based on Proof-of-Work, however, unlike many existing PoW networks that only use the hash power 
-to validate and seal new blocks, our innovative mining algorithms can reuse the hash power 
-to unlock encrypted data and procedures in a controlled manner at the preset timestamp. See details in our [technical paper](https://arxiv.org/abs/2205.09020)
-This dual-purpose mining algorithm is not only securing network safety and data integrity but also serves as a timed-release mechanism. 
-No one, including the originator, can stop releasing the encrypted contents unless all nodes in the network were eliminated. 
-No third party has any type of early access to the encrypted contents. 
-The attacker must utilize hash powers higher than the whole network to prematurely access the encrypted contents 
-so that our system provided unparalleled security and robustness to any centralized TRE services.
+The dTRE technology is enabled by combining the latest confidential computing hardware and blockchain technology. 
+The data is encrypted with a symmetric encryption scheme and sent to random trusted off-chain workers for release. No third party has any early access to the encrypted contents.
+The attacker must defeat security hardware to access encrypted data with unknown values. Therefore, we can build a decentralized reputation on the Polkadot-based blockchain ecosystem and security hardware backed by semiconductor giants (e. g. Intel, AMD and ARM). The dTRE technology provides content creators with a critical tool to monetize their reputation.
 
 The emerging Web3 technology and Polkadot ecosystems could enable enormous cross-chain applications which need time-sensitive and time-critical encryption of data and procedures, 
-including DAO governance, DeFi applications, decentralized prediction market and strategy market. 
-In many scenarios, the data and procedures are required to be confidential until release time. 
-TREX and its blockchain application programming interface (API) could confidently empower the TRE applications on Web 3, 
-which could also terminate cheap talks and scams on today’s social networks and creates entirely new marketplaces 
-for trading time-sensitive confidential data and procedures with a blockchain-backed reputation system.
+including DAO governance, DeFi applications, decentralized prediction market, and strategy market. The data and procedures must be confidential in many scenarios until release time. TREX and its blockchain application programming interface (API) could confidently empower the TRE applications on Web 3. With a blockchain-backed reputation system, it could terminate cheap talks and scams on today’s social networks and create new marketplaces for trading time-sensitive confidential data and procedures.
 
 The TREX project roots a new family of Web3 applications in the Polkadot ecosystem and is being actively developed with the Substrate framework.
 It aims to support interoperation between parachains in the Polkadot ecosystem and enables a group of new applications with dTRE technology.
 The TREX project was initialized to support an idea that needs immutable open records but is hidden until future events.
-Applications like elections and blind bidding can be done on Web3 without relying on centralized services. Our team believes that TRE 
-will support a new type of Web3 application which accelerates the adoption of Web3.
+Applications like elections and blind bidding can be made on Web3 without dependency on centralized services. Our team believes that TRE will support a new type of Web3 application which accelerates the adoption of Web3.
 
 ### Project Details
 
-Our targeted clients are strategy markets and AI trading bot markets that need to build up their decentralized reputations (e. g. [tipranks.com](https://www.tipranks.com)).
+Our targeted clients are strategy markets and AI trading bot markets that need to build up their decentralized reputations (e. g. [tipranks.com](https://www.tipranks.com)). The early encrypted data as a part of the on-chain extrinsic data is immutable; the released early data can help a creator to establish its Web3 reputation. In addition, the delayed release mechanism can effectively monetize the client's Web3 reputation by selling the early access privilege to subscribers.
 
 - Core functionality in TREX network
-![block structure](https://sat02pap001files.storage.live.com/y4mC7nJfXtBtZL0isK0XVuSb1OBHB6ULIEzLVL0nwHNrg81CyWAwQ-VL5bslF7EiXILsdqH6H0bNJ2zs8kVfOn7hYkdEt60_gZvt83AKPR-cHQLgaMvMA-7mpy7OUXcfHReEdp6V5g-xJd-jP7G5QjQBiMOwM0jEIbfnq1dYBNMYSGfo5K8TZeviKs75_qekYmL?width=877&height=428&cropmode=none)
 
+The TREX network is derived from confidential computing technology that creates a trusted execution environment (TEE) inside an off-chain worker so that a key to access the confidential data can be temporarily held inside the off-chain worker's enclave and released in a pre-determined future.
+Besides the regular validator nodes, as in any other substrate-based blockchain network, there is a special kind of "keyholder" node that host off-chain workers in TEEs.
 
-The core functionality in the TREX network includes a deterministic algorithm to generate an infinite series of public keys to encrypt data for future blocks (each block also has a timestamp that labels paired future events) and a PoW mining algorithm that serves dual purposes: seal a new block and release encrypted data corresponding to this new block. As shown in the above figure, similar to Bitcoin or any PoW type of blockchain, the hash value of the previous block will be included in the header data of the block, a design that guarantees that all historical data and transaction records cannot be changed because any changes to the earlier block data will cause the new blockchain to have to recalculate the nonce value, that is, rework. 
+![TREX arch](https://sat02pap001files.storage.live.com/y4mCxh3v92-Q3SeR5Le00nfK2p3UqwRvTLNymJrm1VPHO6pyaC3h4g6IjJ3WeSznwK4llcxYDwVjwTuNSYcG31wx7gbuFPcVA8QZL2nACTkJ4SylngdwUusZKr95ThUv7jCwUaqIIXcCKgYKsmoIRrL-sb3Db3sq-hwkTX-CRxhluIuCYcSh0mMzHtpsLG-AySK?width=947&height=685&cropmode=none)
 
-Each block header contains the corresponding public key to encrypt the hidden message, which could be structured data or serialized procedures, to be released in the current block time. The public key and private key are based on a discrete logarithm problem (DLP) cryptosystem. For example, if a user intends to release a message in the future, the message may be encrypted by the corresponding public key. If the publisher of the hidden message wants to make the news immutable and publicly available after the release time (e. g. electronic voting), the message may be written into the blockchain as part of immutable extrinsic metadata, as shown in the above figure.
+As shown in the above figure, a client machine can encrypt its data with a symmetric key and send the data to the current block as a part of extrinsic. To unlock the encrypted data in the future, the client machine must shield the key in use and send the shielded key to the off-chain workers inside the TEE. The client needs to get remote attestation before trusting the off-chain worker to protect the key from third-party accesses. The client can use [Integritee Network](https://integritee.network) for a decentralized attestation service, or it may use some cloud-based remote attestation services (e. g. [Azure Confidential Computing Service](https://azure.microsoft.com/en-us/solutions/confidential-compute/)). [Integritee Network](https://integritee.network) is a successful pioneer developed on the Polkadot ecosystem and provides a confidential computing platform and infrastructure solutions on Web3.
 
-Since the block time can be feedback controlled and stabilized, we can estimate the corresponding block height for the timestamp of a future event, the corresponding public key can be used to encrypt the message to be released in the future event. Therefore, once the corresponding private key were found by the future miner, the message is also released.
-
-![core function](https://sat02pap001files.storage.live.com/y4mbbm4Uv6wQNS3zVzdrNF3c1kP22J47ApRc3XJyfDWnOYxMIumA4ArzqSNPjMsdwNS1SxEg25Fpx7CR5VdhJjgFBsju_derbYKDLbG_RkZNcel63uYzqMWau1BRhUpdKm73wg99zwZ4XwDEMVPtTcbnW5fU71u2DHt6QF7PFznIMcuB7nkD8aqDWG19sBorQRM?width=910&height=308&cropmode=none)
-
-As shown in the above figure, when a miner finds a nonce value that meets the TREX PoW consensus, it can seal a new block and submit it to peer nodes for verification. While the new block is sealed and verified, based on the mining 
-results of current work, a corresponding private key could also be derived from the new nonce value which could be used to decrypt early encrypted data and release data to the general public.
-
-The algorithms and consensus were detailed in our [technical paper](https://arxiv.org/abs/2205.09020).
-
-TREX algorithm provides 3 major advantages and innovations compared to early works
-
-  1. Designed a dual-purpose mining algorithm that reduces the time complexity to solve the puzzle lock from O(2^n) to O(2^(n/2)), significantly improving the encryption strength.
-  2. Integrated an advanced feedback algorithm to reduce 90% block time fluctuation and provide accurate release time.
-  3. Designed a distributed mining pool architecture and algorithms that enable linear parallel performance gains to defend against hash attacks from high-performance cluster computers.
+Inside the protected enclave, a separated and encrypted region for code and data, the key for releasing the on-chain data is inserted into a priority queue based on its release time. Once the release time is approaching its nearest block, the off-chain worker would release the key from the TEE and expose it to the public. Once the key is released, the block index engine can discover the early encrypted data and decrypt it so it can be accessed publicly. The off-chain worker cannot manipulate the release schedule of keys since decentralized services attested the executable, and it has no access to the shielded data in use.
 
 - TREX token economy
 
-The TREX network is comprised of different types of entities including third-party developers, nodes, archivers and users. A simple example of core functionality in the TREX network.
-![example figure](https://sat02pap001files.storage.live.com/y4m-6fn2EHbIEctqPi4P_8PI0jc8B0HaImf_-065KvTi7GH_A9LUzkFnD5gfszLtS9GQujRDwQanrz_WihcMZvj8s6AhATo0xDIY_hQlDJbUdVNgT_qMJaE-N1k5RC5vEBXStgqTOypL9ql-FbjjDXSiDtNBqBeimm_7CBMsqqPHegLLgnu9mH_uCHHGEPhh6aR?width=662&height=586&cropmode=none)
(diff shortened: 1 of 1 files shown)


yangfh2004:
Hi,
It has been 10 days, no one ever leaves any comments on my PR. Please let me know if there is anything wrong with it or if it is not appropriate for the grant application. Thank you.

takahser:
@yangfh2004 sorry for the delay here. We've experienced a lot of submissions lately and are a bit behind in our backlog. I'm going to give you feedback as soon as possible.

applications/TREX_Network.md
```diff
@@ -0,0 +1,284 @@
...
+
+- Core functionality in TREX network
+![block structure](https://sat02pap001files.storage.live.com/y4mC7nJfXtBtZL0isK0XVuSb1OBHB6ULIEzLVL0nwHNrg81CyWAwQ-VL5bslF7EiXILsdqH6H0bNJ2zs8kVfOn7hYkdEt60_gZvt83AKPR-cHQLgaMvMA-7mpy7OUXcfHReEdp6V5g-xJd-jP7G5QjQBiMOwM0jEIbfnq1dYBNMYSGfo5K8TZeviKs75_qekYmL?width=877&height=428&cropmode=none)
+
+
+The core functionality in the TREX network includes a deterministic algorithm to generate an infinite series of public keys to encrypt data for future blocks (each block also has a timestamp that labels paired future events) and a PoW mining algorithm that serves dual purposes: seal a new block and release encrypted data corresponding to this new block. As shown in the above figure, similar to Bitcoin or any PoW type of blockchain, the hash value of the previous block will be included in the header data of the block, a design that guarantees that all historical data and transaction records cannot be changed because any changes to the earlier block data will cause the new blockchain to have to recalculate the nonce value, that is, rework.
```
takahser, line 51: Can you explain what in detail you mean by:

> seal a new block

applications/TREX_Network.md
```diff
@@ -0,0 +1,284 @@
...
+  2. Our [TREX blockchain index engine](https://github.com/NexTokenTech/trex-archive) is based on the Substrate Archive. It releases timed-release encrypted data and archives dTRE data into a PostgreSQL database.
+  3. For applications on the TREX network, we provide solutions including an [API library](https://github.com/NexTokenTech/trex-subxt) to our business partners who may use it to insert dTRE data on the TREX network and retrieve data with the TREX blockchain index engine.
+  4. The DevOps tool uses [Terraform](https://www.terraform.io) to automate the deployment of the TREX infrastructure with other backend services on cloud computing platforms so that our targeted clients can easily integrate the dTRE technology with their existing services.
+
+- Documentation of core components, protocols, architecture, etc. to be deployed
+
+Our white paper is our core documentation.
+
+<iframe src="https://onedrive.live.com/embed?cid=BEF98A7CB59E1060&resid=BEF98A7CB59E1060%21106391&authkey=AGFFIH66aCaJ9Yw&em=2" width="800" height="400" frameborder="0" scrolling="no"></iframe>
```
takahser, line 98: `iframe`s are [not supported](https://github.github.com/gfm/#disallowed-raw-html-extension-) by GitHub markdown, could you remove it and maybe replace it with a normal link?

applications/TREX_Network.md
```diff
@@ -0,0 +1,284 @@
...
+
+| Number | Deliverable | Specification |
+| -----: | ----------- | ------------- |
+| 0a. | License | MIT License
+| 0b. | Documentation | We will provide both **inline documentation** of the code and a basic **tutorial** that explains how a parachain can route a dTRE message to the TREX network.|
+| 0c. | Testing Guide | Core functions will be fully covered by unit tests to ensure functionality and robustness. In the guide, we will describe how to run these tests. |
+| 0d. | Docker | We will provide a Dockerfile(s) that can be used to test all the functionality delivered with this milestone. |
+| 0e. | Article | We will publish an online blog that explains how this module works as described in the following figure.
+| 1. | Substrate module: XCMP | Provide run-time pallet to process cross-chain requests to put dTRE data on the TREX blockchain.
+| 2. | TREX node update | The TREX node is updated to provide any necessary low-level supports for XCMP protocal and cross-chain dTRE data.
```
takahser, line 256:
```suggestion
| 2. | TREX node update | The TREX node is updated to provide any necessary low-level support for XCMP protocol and cross-chain dTRE data.
```

applications/TREX_Network.md
```diff
@@ -0,0 +1,284 @@
...
+- **Costs:** 20,000 USD
+
+| Number | Deliverable | Specification |
+| -----: | ----------- | ------------- |
+| 0a. | License | MIT License
+| 0b. | Documentation | We will provide both **inline documentation** of the code and a basic **tutorial** that explains how a content creator can generate tradable content and a user can gain early access to this content with a fee. |
+| 0c. | Testing Guide | Core functions will be fully covered by unit tests to ensure functionality and robustness. In the guide, we will describe how to run these tests. |
+| 0d. | Docker | We will provide a Dockerfile(s) that can be used to test all the functionality delivered with this milestone. |
+| 0e. | Article | We will publish an online blog that explains how this module works as described in the following figure.
+| 1. | TREX mining server | A TREX node which host a mining pool and access connection from remote miners and send rewards to miners. To protect the consensus against attacks from cluter computers, the mining pool needs to host a large hash table so that the mining efficiency is close to the cluster architecture coordinated by a center.
```
takahser, line 271:
```suggestion
| 1. | TREX mining server | A TREX node which host a mining pool and access connection from remote miners and send rewards to miners. To protect the consensus against attacks from computer clusters, the mining pool needs to host a large hash table so that the mining efficiency is close to the cluster architecture coordinated by a center.
```

takahser (changes requested):
@juntao thanks for submitting your grant application.

I have a few questions:

> The attacker must utilize hash powers higher than the whole network to prematurely access the encrypted contents so that our system provided unparalleled security and robustness to any centralized TRE services.

Does this mean, 51% attacks are not possible with this PoW algorithm?

Also, I was wondering if this PoW algorithm is going to be ASIC-resistant?

Finally, feel free to take a look at my inline questions and suggestions I added.

applications/TREX_Network.md
yangfh2004: Hi, I think what I descrbied is not about network consensus but encryption strength. The timed-release encryption is basically a crypto system which allow user to encrpyt data and decrypt it automatically in predetermined future. What I mean is you need to get more computing power than the network to get earlier access independently to the encrypted contents than everyone else in the network. This is not about to take over the blockchain but decrypt the encrypted contents. For the consensus, you need to have 51% or more hash power to take the blockchain, which is the same as other PoW blockchains.

For the ASIC-resistancy, it is about its hash function. Our algorithm can interface with different hash function, if some hash function have ASIC to accelerate, we could change to other hash function which is ASIC-resistant. The hash function is essentially providing peusdo-randomness in building the graph and finding the collision to solve the puzzle (which is the encryption scheme), thus, many hash function works with our algorithm framework (e. g. SHA256, BLAKE3, etc). I am working on some documentation to explain more about the algorithm. I could also explain it to you if you are interested.

burdges:
We should reject anything PoW imho, as broadly polkadot is too optimized for proof-of-work to be useful anywhere, but specifically here: 

I'm dubious that Pollard's rho method yields predictable timings, due to the existence of parallelized algorithms.  ASICs always represent a threat too.

Isogeny VDFs are extremely expensive but if they work then delay encryption by Luca and I achieves better, under stronger sequential computational assumptions:  https://eprint.iacr.org/2020/638

If isogney VDFs are unrealistic then one could improve upon DFinity's DKG, which yields IBE and thus delay encrpytion in a Byzantine threat model, but annoying liveness assumptions on the DKG.

It's also possible variations on https://github.com/kobigurk/aggregatable-dkg achieves some similar time release encrpytion, which improves upon Dfinity's DKG for liveness, but yields less than IBE which complicates decryption liveness.

Any exotic delay encryption related funding we do should be either DKGs or isogenies, maybe in collaboration with E.F. for isogenies.

We also have many validators in europe where power has become staggeringly expensive.

yangfh2004:
> We should reject anything proof-of-work imho, as broadly polkadot is too optimized for proof-of-work to be useful anywhere, but specifically here:
> 
> I'm dubious that Pollard's rho method yields predictable timings, due to the existence of parallelized algorithms. ASICs are always represent a threat too.
> 
> Isogeny VDFs are extremely expensive, due to ASIC boards needing considerable RAM and many SSD controllers, but if they work then delay encryption by Luca and I achieves better, under stronger sequential computational assumptions: https://eprint.iacr.org/2020/638
> 
> If isogney VDFs are unrealistic then one could improve upon DFinity's DKG, which yields threshold IBE and thus delay encrpytion in a Byzantine threat model, but annoying liveness assumptions on the DKG.
> 
> It's also possible variations on https://github.com/kobigurk/aggregatable-dkg achieves some similar time release encrpytion, which improves upon Dfinity's DKG for liveness, but yields less than IBE which complicates decryption liveness.
> 
> Any exotic delay encryption related funding we do should probably be either DKGs plus related encryption schemes, or else isogenies, maybe in collaboration with E.F. for isogenies.
> 
> We also have many nodes in europe where power has become staggeringly expensive.

Hi, parallel algorithms can be dealt with a mining pool so that all nodes in the same pool can do work in parallel. ASICs do not exist so far and it is too early to talk about. About the energy concern, I agree with you that energy could be very expensive in Europe but this is not only for Europe, the internet covers the whole world.

Our parachain has nothing to compete with Polkadot but join the ecosystem to do some extra work. if PoW is solely for verification of transactions. I agree with you, PoW is useless here. But in our protocol, PoW is essential to decrypt some data with timed delay. We simply have to use computing power to solve puzzles.

DKG may work for decentralized authorization but how do you generate keys for a distant future many blocks away without spending too much time/energy on communication and verification? Our PoW protocol can generate a future key with a deterministic algorithm independently without any communication or verification. Our system is not exotic but simple and efficient.

I respect your point of view and I am glad to read and learn your insights. Thank you.

burdges:
ICP employs resharing in their DKG, so in theory they can generate arbitrarily far into the future.  It's incurs some liveness concerns and  complexity like I said above, but it works in a sane-ish threat model.

All proof-of-work deployments appear [insecure](https://www.crypto51.app/), except perhaps bitcoin, so really mining pools cannot avoid the inherent weakness of using parallel algorithms.  Also, even bitcoin really only achieves "security" via ASICs acting like de facto stake.

Aside from the work cost itself, proof-of-work opportunity costs caused by being unable to optimally use resources.  At present proof-of-work parachains would miss most of their parachain slots.  We'll land parathreads eventually so then I guess miners could buy parathread blocks whenever they mine a block, but internally the parachain remains slow & inefficient.  As an aside, there are proof-of-stake schemes with this problem too like ouroboros praos and fantômette.

We'll eventually have direct parachain-to-parachain messaging called XCMP but it'll break for proof-of-work parachains, isolating them somewhat, although our existing messaging schemes work.

Anyways...

Why do you want encryption so far into the future?  An MEV defense only need only very short-term encryption, for which more methods exist.  Are you assuming some large adversarial advantage like VDFs do?

yangfh2004:
> ICP employs resharing in their DKG, so in theory they can generate arbitrarily far into the future. It's incurs some liveness concerns and complexity like I said above, but it works in a sane-ish threat model.
> 
> All proof-of-work deployments appear [insecure](https://www.crypto51.app/), except perhaps bitcoin, so really mining pools cannot avoid the inherent weakness of using parallel algorithms. Also, even bitcoin really only achieves "security" via ASICs acting like de facto stake.
> 
> Aside from the work cost itself, proof-of-work incurs opportunity costs caused by being unable to optimally use resources. At present proof-of-work parachains would miss most of their parachain slots. We'll land parathreads eventually so then I guess miners could buy parathread blocks whenever they mine a block, but internally the parachain remains slow & inefficient. As an aside, there are proof-of-stake schemes with this problem too like ouroboros praos and fantômette.
> 
> We'll eventually have direct parachain-to-parachain messaging called XCMP but it'll break for proof-of-work parachains, isolating them somewhat, although our existing messaging schemes work.
> 
> Anyways...
> 
> Why do you want encryption so far into the future? An MEV defense only need only very short-term encryption, for which more methods exist. Are you assuming some large adversarial advantage like VDFs do?

Thanks a lot for your informative comments.
I am not a crypto researcher but a software engineer. Our team's ultimate goal is to build a timed-release encryption system to help creators build web3 reputations and monetize them by selling the early access privilege. For example, if you can predict the weather in 5 days and you have reputation in weather prediction. You want to encrypt your prediction to be released after 5 days. And buyers can pay you to access your prediction before the event happens. More details is in our pitch deck. I can send it to you or include it in this application.
My element account is @yangfh2004:matrix.org
PoW is not our first choice, but our current consensus relies on it. I somehow agree with you that most of the current PoW network is "insecure". My only argument is that the PoW works as long as the value it holds is less that the cost of attack. A network like Bitcoin requires a huge amount of power to secure itself, but some less valuable application does not need that much power unless you assume that someone would like to damage others without gains. 
We are still in the very early phase. As you mentioned that ICP may have some technology of DKG, I will look into that if you allow me to spend a few more days and re-consider our consensus design and plans. I will update my applications after my research.
Again, thank you very much!

burdges:
I see..  You want incorrect predictions revealed then, while whoever makes incorrect prediction wants them hidden.

ICP's code is closed source, so not too helpful.  There are simpler DKGs like drand that work too, but they've worse liveness. 

I've no idea about the simplest scheme here, as numerous prospects present themselves, but it always helps if you've some idea how valuable everything is.  Is some black mark for not revealing enough?  Or you think reveals wind up time sensitive?

Intel SGX might even suffice if reputation wins up not too valuable and if you avoid future directions for likely SGX breaks ala speculative execution, etc.

yangfh2004:
> I see.. You want incorrect predictions revealed then, while whoever makes incorrect prediction wants them hidden.
> 
> ICP's code is closed source, so not too helpful. There are simpler DKGs like drand that work too, but they've worse liveness.
> 
> I've no idea about the simplest scheme here, as numerous prospects present themselves, but it always helps if you've some idea how valuable everything is. Is some black mark for not revealing enough? Or you think reveals wind up time sensitive?
> 
> Intel SGX might even suffice if reputation wins up not too valuable and if you avoid future directions for likely SGX breaks ala speculative execution, etc.

I think Intel SGX might work for us so that we can get rid of PoW. I think if we put extrinsic contains time-delayed secrets into a separate memory pool safeguarded by SGX. It will be put on blockchains when it arrives on its release time. Before that, it remains secret inside the enclave. Thanks a lot for your advice. I will revise my grant application and update it in a few days!

burdges:
I looked again at your paper https://arxiv.org/pdf/2205.09020.pdf but it appears each block announces a new public key, which the next block decrypts, so you do not have long delays here either, no? 

I'm afraid RSA VDFs remain the delay primitive that's anywhere near secure-ish deployment, but they stalled when the trusted setup proved too difficult for the Ethereum foundation, and wont support delay encryption.

Isogeny VDFs could be securely deployed, and they've a tractable trusted setup, but they require much larger more complex ASICs with considerable RAM and many SSD controllers with which they read parameters.  Also, they only deliver fairly short runtimes because evaluators need 14 terabytes RAM speed storage per hour.  All those SSD controllers permit reading parameters in parallel to achieve RAM-like speeds reading from SSDs.

yangfh2004:
> 

Hi, thanks for your reading. The new block does not announce the public key. The public key is generated by a deterministic algorithm using the previous public key as a seed. The first public key is hard-coded in the genesis block. Thus, theoretically a near-endless list of public keys can be generated independently and assigned to each future block in a sequence. 

You cannot pre-mine a future block since it will require a hash of the previous block header. But you can decrypt the content by other meanings (since you know the public key). Thus, the encryption strength of each public key will not make the miner work longer than each block time (I think that is what you mean in your question). If you want to enhance the encryption strength, you can encrypt the content by many layers using adjacent public keys. 

My business model assumes that each encrypted content has relatively small values, and the attack cannot easily estimate the value of each encrypted data before it has been released. They have to spend relatively large computing power to see the encrypted content before the pre-determined release time. But they may get incorrect information or just junk. They better use the same power to mine since it gives them a stable return.

yangfh2004:
Hi, I just pushed a revision to our grant application to get rid of PoW consensus. We decided using confidential computing technology (e. g. Intel SGX) to implement our dTRE platform. It is a significant change to us. I think @burdges suggestions on using SGX is really helpful to us and provides us a new field to explore different solutions. I think the SGX technology fits our application very well since single peice data does not have very high value. The existing Integritee Network project demonstrates the possibility of using this technology and provides us a good reference. Please review our revised application and leave comments.
Please let me know if I need to squash two commits into a single one or leave it as is. Thank you.

burdges:
Just be careful, SGX gets broken periodically, so ideally you'll want some "defense in depth" there.

You could try key splitting, but you'll want good randomness to stop [rewinds](https://docs.conclave.net/security.html#mementorewindreplay-attacks), and randomness gets [tricky](https://www.theregister.com/2020/06/10/intel_patches_sgx_again/).

It's likely drand does what you want eventually too, in a threshold threat model, but they'd take a couple years I think.

Noc2 (changes requested):
Thanks for the application, all the updates, and sorry again for the delay here. In general, given that this is a level 1 grant, I think the committee is very interested in supporting these experiments, and the development of PoCs. And it looks like you already put a lot of effort into this. But I have two more comments: 

- Could you remove Integritee network from the grant application and rather simply focus on the development of open-source software? When it comes to grants we want to be as objective as possible and not support a specific parachain? 
- Regarding SGX, you should be aware of the following: https://eric-diehl.com/intel-sgx-is-dead/, which is obviously a huge problem for all the SGX-based projects out there.

burdges:
I dislike defending SGX but afaik SGX is only dead for consumer CPUs, no?  I thought server CPUs like Xeons still have SGX?

yangfh2004:
> Thanks for the application, all the updates, and sorry again for the delay here. In general, given that this is a level 1 grant, I think the committee is very interested in supporting these experiments, and the development of PoCs. And it looks like you already put a lot of effort into this. But I have two more comments:
> 
> * Could you remove Integritee network from the grant application and rather simply focus on the development of open-source software? When it comes to grants we want to be as objective as possible and not support a specific parachain?
> * Regarding SGX, you should be aware of the following: https://eric-diehl.com/intel-sgx-is-dead/, which is obviously a huge problem for all the SGX-based projects out there.

Thanks a lot for your comments. I will remove their name, "Integritee Network" from the application. The reason that I mentioned them is that they are providing a decentralized remote attestation service as a part of the Polkadot Ecosystem. We will rely on their remote attestation service, at least in our early phase.

For SGX, I did some research, and I agree with @burdges.  I believe Intel phases out this technology from consumer CPUs since not many important consumer application needs it besides playing blue-ray discs on your PC. And people are mostly watching streaming videos today, and the most computer doesn't even have a disc drive. but Xeon scalable processors still have them with their latest generation. Most cloud computing platforms (e. g. AWS, and Azure) use Xeon scalable processors and support SGX very well. You may easily deploy a container on a virtual machine that runs with Xeon processors. And you can still find a lot of consumer-grade processors (6th - 10th gen core processors) with SGX supports if you want to build a cheap server in your garage.

I think Intel will eventually replace SGX with TDX technology in the next few years, as I mentioned in our future plan. We can then update some off-chain worker code to support the latest generation of hardware. Since it is an off-chain worker. The older generation hardware is still being supported, and there is no need to change any code for on-chain operations. Many other chip makers like (AMD, Nvidia, and ARM) are also working on similar technology. I think we will have more and more choices to support our network in the long run, but we cannot just sit there and wait for everything ready. If I have a chance to receive this grant, I will try to go to the next levels and keep working in this area for the next few years.

yangfh2004:
> Just be careful, SGX gets broken periodically, so ideally you'll want some "defense in depth" there.
> 
> You could try key splitting, but you'll want good randomness to stop [rewinds](https://docs.conclave.net/security.html#mementorewindreplay-attacks), and randomness gets [tricky](https://www.theregister.com/2020/06/10/intel_patches_sgx_again/).
> 
> It's likely drand does what you want eventually too, in a threshold threat model, but they'd take a couple years I think.

Hi,
Thanks a lot for your comments. Key splitting is definitely helpful to secure the key from early release in case of security attacks. I will include it in the application. I think the "rewind" attacks may only delay the release of the key or duplicate it. As long as the key is not released before the schedule, it would not be a serious issue for our application. I will add description of how we avoid relying on the system clock to prevent the key from releasing earlier than the schedule.

burdges:
If you've a delay within one epoch, then you could enforce x relay chain blocks as the delay by checking the new light client proofs provided by BEEFY.  If you've a delay longer than one epoch, then you can check those proofs for transitioning validator sets too.  

In either case, the encryption specifies the light client commitment to the relay chain validator set for the epoch when encryption is done, which saves us caring about the relay chain history before that point.  An SGX enclave can verify these light client proofs, and compute the decryption key if correct.  All this gives us threshold model of the relay chain, despite the relay chain not providing threshold signing.

There is only one secret being protected by SGX here, a secret key corresponding to a public key to which all messages are encrypted.  As IBE is not required, you could use any key exchange you desire here, so you could pick a key exchange friendly to both key splitting and constant-time execution.

You could've a merkle tree of these secret keys with one per future polakdot epoch, so then these SGX enclaves would not decrypt using a particular key unless they observe the BEEFY signatures, which never exist until the correct epoch.  It's maybe possible to row hammer the enclave into decrypting early, but this sounds like a harder attack, and additional defenses-in-depth apply.

You must do a trusted setup for these secret keys, but this could be done in a threshold way too.  I do not know how secret data is prepared for an SGX enclave, but this maybe the weakest point if the SGX enclave's own decryption key is vulnerable. 

----

Aside from SGX, then you could switch between threshold models using similar techniques, like assuming >1/2 parachain decryptor nodes are honest, then decryption only happens in 2/3rd of relay chain validators have signed in BEEFY.

yangfh2004:
> If you've a delay within one epoch, then you could enforce x relay chain blocks as the delay by checking the new light client proofs provided by BEEFY.  If you've a delay longer than one epoch, then you can check those proofs for transitioning validator sets too.  
> 
> 
> 
> In either case, the encryption specifies the light client commitment to the relay chain validator set for the epoch when encryption is done, which saves us caring about the relay chain history before that point.  An SGX enclave can verify these light client proofs, and compute the decryption key if correct.  All this gives us threshold model of the relay chain, despite the relay chain not providing threshold signing.
> 
> 
> 
> There is only one secret being protected by SGX here, a secret key corresponding to a public key to which all messages are encrypted.  As IBE is not required, you could use any key exchange you desire here, so you could pick a key exchange friendly to both key splitting and constant-time execution.
> 
> 
> 
> You could've a merkle tree of these secret keys with one per future polakdot epoch, so then these SGX enclaves would not decrypt using a particular key unless they observe the BEEFY signatures, which never exist until the correct epoch.  It's maybe possible to row hammer the enclave into decrypting early, but this sounds like a harder attack, and additional defenses-in-depth apply.
> 
> 
> 
> You must do a trusted setup for these secret keys, but this could be done in a threshold way too.  I do not know how secret data is prepared for an SGX enclave, but this maybe the weakest point if the SGX enclave's own decryption key is vulnerable. 
> 
> 
> 
> ----
> 
> 
> 
> Aside from SGX, then you could switch between threshold models using similar techniques, like assuming >1/2 parachain decryptor nodes are honest, then decryption only happens in 2/3rd of relay chain validators have signed in BEEFY. 
> 
> 

Thanks a lot for your suggestions. I need to digest your post and I will reply after this weekend.

Noc2 (approved):
Thanks for the update. I'm already happy to go ahead with this (given the amount of work you already put into this, as well as that it's only a level 1 grant) and share it with the rest of the team.

yangfh2004:
> Thanks for the update. I'm already happy to go ahead with this (given the amount of work you already put into this, as well as that it's only a level 1 grant) and share it with the rest of the team.

Thanks a lot for your approval. @burdges has given us a lot of great comments, and we are working on integrating his suggestions into our next-level application. For the current level, we are working on the delivery of a minimum viable product based on Intel SGX technology and gradually enhancing security by integrating all necessary encryption schemes and modules.

takahser (approved):

yangfh2004:
> 

@burdges  After digesting your comments many times and spending hours reading materials about BEEFY protocol, I believe I got your idea, and I think your idea enlighted our project. 
Originally, I thought I might use some trusted time service from Intel to control the release time or some specific instructions for timing (which might be inaccurate). 
As you mentioned, I can use the BEEFY signature to control the release of private keys based on the finality block height. Since it is very hard to fake more than 2/3 BEEFY finality proof from the Polkadot relay chain, the BEEFY signature becomes a trustworthy decentralized timestamp generated from the Polkadot network. If I understand it correctly, I think it is a much better solution. 
Moreover, if I understand your comments correctly, you suggested that we may generate all future secret keys within the enclave by using the hash of a parent key. All we need to do is to set up the first key for the genesis block in a threshold way.
I am not gonna worry about the row hammer or any other hardware hack for now. Thanks again for your comments. I will build a proof of concept.

keeganquigley:
Hi @yangfh2004 can you give us an update on your progress?

yangfh2004:
Yes, I just submitted PR for milestone deliveries a few days ago. We developed a substrate based node, an off-chain worker and CLI tools for testing and configuring.

yangfh2004:
https://github.com/w3f/Grant-Milestone-Delivery/pull/724

keeganquigley:
@yangfh2004 awesome thank you!

PR #1207, opened 2022-10-10
Change digitanativeinc to standardweb3

hskang9:
I would like to announce that we have changed our github organization id from digitalnativeinc to standardweb3.

I think CI is pointing to the previous organization here too.

https://github.com/w3f/Grant-Milestone-Delivery/pull/576


Noc2 (approved):
Thanks for the update. I will ping the comms team as well.

hskang9:
Just to note, the link is not changed.

PR #1269, opened 2022-11-10
Replace emojis in accepted_grant_applications.md for docusaurus

semuelle:


Noc2 (changes requested):
Is there also an option for a box that has no cross in it?

semuelle:
☐

Like so?

Noc2 (approved):
I updated it. Feel free to merge it, if you are happy with it.

PR #1295, opened 2022-11-22
Update index.md

Noc2:
Fix intro text


semuelle:
Is that the same change as in the master branch? In that case, I'd prefer to do a manual merge of everything that's already on master soon-ish.

Noc2:
But in the master branch, this is only changed in the readme, not in the index.md

semuelle:
Right! In that case, proceed. :)

PR #1312, opened 2022-11-28
initial commit

jonasW3F:
This is an RFP to find an implementer of the validator selection algorithm as described [here](https://papers.ssrn.com/sol3/papers.cfm?abstract_id=4253515). The RFP aims to get an implementation that can be used by various nomination services around the ecosystem.


Noc2 (approved):
Thanks for creating the RFP. I’m happy to already merged, and we will continue to improve it a little bit before we share it publicly (for example, adding the status, etc. )

PR #1329, opened 2022-12-02
Guidelines adjustments for better accommodate research grant applications

dsm-w3f:
- Split the requested information for a grant application into two kinds of projects: software development and research.
- Add Google Scholar profiles for research grants.
- Add Research as a category of project in the guidelines.


Noc2 (approved):
Looks good. I’m happy to go ahead already and merge it since it is definitely better than the current status. We might want to think about having a separate template for it. The current milestone deliveries etc, are also pretty development-focused.

PR #1341, opened 2022-12-07
Links to the grant contracts in the applications index

randombishop:
Add Grant column and populate it using heuristic script


Noc2 (changes requested):
That is great. Is it possible to link to the merged applications instead of the PR. Ideally, in a way that also works on the website: https://w3f.github.io/Grants-Program/Applications

PR #1358, opened 2022-12-13
Create Application Template for Research Grants

dsm-w3f:
- Add new application template for research grants
- Remove research information from software development application template
- Add reference to research grant on Readme.md.


Noc2 (approved):
Thanks. I think this will already be really helpful.

PR #1377, opened 2022-12-19
Update SaaS3.md - fix website

0xCaso:
Closing `` tags (website was broken)


Noc2 (approved):
@0xCaso Happy to merge it, but does this actually solve the issue, or do you need to use markdown for images?

0xCaso:
@Noc2 It partially solved the issue; actually, we should ensure that every tag is closed otherwise the website will break.

PR #1404, opened 2022-12-30
Amend Ink Smart Contracts Wizard

avirajkhare00:
There are a few changes in the grant:
 - Creation of CLI tool for code generation instead of Web App.
 - Reduced price of M2 by $1,000
 - Introduced M3
 - Using OpenBrush Smart Contracts instead of Ink Examples.
 - Change in BTC address.


Noc2 (changes requested):
Thanks for the amendment. I have a couple of questions: 
- Could you specify what exactly you mean by components? 
- What functionality will the CLI support? 
- Regarding the creation of the formula, are you referring to the following process: https://docs.brew.sh/Formula-Cookbook

avirajkhare00:
Hey @Noc2 @0xCaso Happy new year.

> Could you specify what exactly you mean by components?

Using word Template would be much better here.

For example, here is a smart contract converted to a template:

```
#![cfg_attr(not(feature = "std"), no_std)]
#![feature(min_specialization)]
        
#[openbrush::contract]
pub mod my_psp22 {
    
    // imports from openbrush
	use openbrush::contracts::psp22::*;
	use openbrush::traits::Storage;

    #[ink(storage)]
    #[derive(Default, Storage)]
    pub struct {token_name} {
    	#[storage_field]
		psp22: psp22::Data,
    }
    
    // Section contains default implementation without any modifications
	impl PSP22 for {token_name} {}
     
    impl {token_name} {
        #[ink(constructor)]
        pub fn new(initial_supply: Balance) -> Self {
            let mut _instance = Self::default();
			_instance._mint_to(_instance.env().caller(), initial_supply).expect("Should mint"); 
			_instance
        }
    }
}
```

Here, when user gives name to the token, using string interpolation, code will replace name of token given by user with the {token_name}. Correct should be, converting smart contract into base templates and it's sub templates like mint block, burn block, etc.

> What functionality will the CLI support?

CLI will support following functionalities:

1. Creation of Flipper Smart contract(No UI). Same as cargo contract new does.
2. Creation of PSP-22, PSP-24, PSP-37
I will add all the commands in the grant application.

> Regarding the creation of the formula, are you referring to the following process: https://docs.brew.sh/Formula-Cookbook

Yes. It will be easy for the users to install the CLI, which is one command away.

0xCaso (changes requested):
Hi @avirajkhare00!

Thanks for the explanation and responses. In my opinion, this grant could be a little bit less expensive. By the way, I think the milestones' objectives are well defined. This is how I'd edit the amendment:
- Milestone 1 price: 3.000 USD;
- Milestone 3 merged into Milestone 2, and for a price of 2.000 USD;

Let me know what do you think about it.

Also, do you think it could be a nice thing to integrate ink-boxes with the CLI? So maybe a user could download a box with it and starting the local development with also the related front-end template?

avirajkhare00:
Hey @0xCaso. First of all, apologies for removing you from the request review. I thought GitHub had the option to have multiple reviewers. For sure we can merge M2 and M3 for the price of $2,000. For Milestone 1, it will have 4 basic templates and up to 20 sub-templates. If you think $3,000 is reasonable, then let's go with it.

0xCaso:
Great, thanks a lot, and no problem. Let's see what @Noc2 thinks about it.

Noc2 (changes requested):
Sounds good to me. Feel free to update it accordingly.

0xCaso (approved):
LGTM

keeganquigley (approved):

PR #1407, opened 2022-12-31
EightFish Application

miketang84:
#### Project Abstract

> Please replace these instructions with a brief description of your project summarising key points (1-2 paragraphs).
>
> If your application is a follow-up to a previous grant, please mention which one in the first line of the abstract and include a link to previous pull requests if applicable.

#### Application Checklist

- [x] A BTC, Ethereum (USDC/DAI) or Polkadot/Kusama (aUSD/USDT) address for the payment of the milestones is provided in the application.
- [ ] I prefer the discussion of this application to take place in a private Element/Matrix channel. My username is: `@daogangtang:matrix.org` (change the homeserver if you use a different one)


Noc2 (changes requested):
Thanks for the application. I have a couple of initial questions/comments:
- Could you be more specific with your milestone deliverables and the specification? "Some tools or scripts to help on code upgrade," for example, is very unspecific and isn't something that we can evaluate since it basically could be anything. 
- You might want to specify the "components" as part of the milestones.
- What exactly do you mean by Optimize, and which new algorithm do you want to implement? 
- How do you make sure the off-chain data is available, and have you considered using IPFS?

miketang84:
Thank you @Noc2 . Will revise this PR and later comment here.

miketang84:
Hi, @Noc2, I have updated this PR upon your 4 review points, please check. 

Regards.

Noc2 (changes requested):
Thanks for the update. I'm generally happy to go ahead with milestones 1 and 2. However, milestone 3 seems to be not very specific at this stage. For example, what do you mean by "Another stuff matters too." and isn't this approach risky? Therefore my recommendation would be to initially only apply for milestones 1 and 2 and potentially later apply for a follow-up grant for milestone 3.

miketang84:
Great appreciated for your kind suggestion. I've moved milestone 3 out of this pr and to future work. Thank you.

Noc2 (dismissed):
Thanks for the quick reply. As promised, I'm happy to go ahead with it and share it with the rest of the team.

dsm-w3f (commented):
Thank you for your application. I was reading the paper to better understand the solution proposed. Initially, I was thinking that was a new decentralized layer on the top of the substrate allowing faster transactions and a better developer/user experience. However, then I started to notice some limitations, such as "There’s no transferable asset on an open web application network." which means that there are no tokens, this is replaced by "credits" that I really don't know what it means. Furthermore, the document says "there is no need to force users to use a web browser wallet plugin to kick off the beginning of the application", hence how the addresses/identity/keys of the user will be held? it will be a native wallet? 

I was thinking that OWA would be a new concept on top of web3, which I think is a good idea. However, after reading maybe it cannot handle all things we have in web3 and it is trying to replace it with new concepts, which I don't think is a good idea. Could elaborate more on that? Would be possible for the cases that OWA cannot handle (such as tokens/assets I think), just delegate to the substrate? This is just to not limit the OWA and create new concepts where already exist some. There is always a risk that users don't understand or want to adopt these new concepts.

miketang84:
Thank you @dsm-w3f for reading the background paper carefully. My explanations as following:

> Thank you for your application. I was reading the paper to better understand the solution proposed. Initially, I was thinking that was a new decentralized layer on the top of the substrate allowing faster transactions and a better developer/user experience. However, then I started to notice some limitations, such as "There’s no transferable asset on an open web application network." which means that there are no tokens, this is replaced by "credits" that I really don't know what it means. 

Actually, there is token in a OWA application/network, this token would be used as the rewards to validators of every new block generation, and used for future govenance of this network.

The reason we would like to eliminate the transferable asset in a OWA network, is to eliminate the dependencies of the PoS colaterals, which is used to keep the network more secure, and further is to reduce the essential number of nodes in this network.

And the credits of a OWA could be either implemented in the off-chain storage, or implemented in the substrate runtime pallet, but remove the function of transfer.

Sorry, the paper on the Medium may be outdated, and I keep adding new content to this paper once I have new meditations. Please refer the laster [google docs: The Road to Open Web](https://docs.google.com/document/d/1XNd7zmXK15ST7LbYSxYhCsSQ9xZgjuv9QbvqNbOdszE/edit?usp=sharing). At the section of this paper, [The Connection between the OWA and the Financial Layer, and Regulation](https://docs.google.com/document/d/1XNd7zmXK15ST7LbYSxYhCsSQ9xZgjuv9QbvqNbOdszE/edit#heading=h.mvfwir2tppb), this is a figure to reveal the notion.

gateways

On the OWA side, we focus on handling the problem of business, and on the financial side, we focus on handling the problem of assets. We believe that a smart contract public chain is the right place to handle the asset and financial biz. Between them, there are gateways. For each public blockchain, it needs a gateway. For example, if we want to do asset exchanges on Moonbeam, we need a OWA-Moonbeam gateway. This gateway is like a cross-chain bridge, but I think it needs only half part features of it.

> Furthermore, the document says "there is no need to force users to use a web browser wallet plugin to kick off the beginning of the application", hence how the addresses/identity/keys of the user will be held? it will be a native wallet?

At the beginning, maybe a OWA end user will use google/github/facebook account to log in this application, and later, depending on the business, when it's the time, this application will bind the Web2 id (a unique string) with a newly created Substrate account, and ask user to store its key in safe place. And the user must use the browser wallet to login into the application to redeem the credits via aboving gateways.

But the user could still use the Web2 oauth method to login for using this application.

> I was thinking that OWA would be a new concept on top of web3, which I think is a good idea. However, after reading maybe it cannot handle all things we have in web3 and it is trying to replace it with new concepts, which I don't think is a good idea. Could elaborate more on that? Would be possible for the cases that OWA cannot handle (such as tokens/assets I think), just delegate to the substrate? 

Yes, the OWA solution is not to solve all problems Web3 will handle, its focus is on the business logic part, and the asset related issues are better to be put into smart contract public blockchains. The best case is they can cooperate together. The OWA part has higher performance, better programming interface and better UI/UE. The asset/financial layer has higher security insurance to keep the user's asset more safe.

> This is just to not limit the OWA and create new concepts where already exist some. There is always a risk that users don't understand or want to adopt these new concepts.

For this point, I think for existing Web3 developers, the OWA is a relative new concept. But for the traditional Web2 developers, which is 100 times more that Web3's, the OWA is a very easy concept, since the EightFish we are working on is a framework very like the Web MVC development framework familiar by those developers. We can attract them into the Web3 space.

And for the end users, I think they have no necessaries to understand these concepts, just use the application and feel happy, that's the most important thing. :)

Wish I post my notions clearly.

Regards.

dsm-w3f (dismissed):
@miketang84 thank you for your answers. I think now I understand the idea, it is initially more a decentralized web2 app with an open data layer supported by substrate than a new layer for a web3 application. I think is a good starting point. However, please notice that some fundamental properties of web3 could be lost in this schema, such as ownership of data since the app will generate the keys and maybe store them. I would like more if the app worked with non-custodial keys. Binding those keys with web2 OAuth identities could be a promising solution for users to migrate from web2 to OWA or even web3. One more thing, when you say bridges please notice that substrate supports XCM, which mainly eliminates the need for parachain bridges. Please consider using XCM instead of bridges in your solution. 

I understand that the solution is in an initial phase and it is very difficult to envision all architectural parts and solutions upfront. There is an actual necessity to migrate users/developers from web2 to web3 in a frictionless way. I think this project fits well in trying to solve this problem. I hope that the limitations mentioned be present only in an initial solution and that you can find a solution in the long term for this project to be a full web2 to web3 adapter/runtime.

I'm happy that you have selected the substrate to develop this project and happy to support it. Good luck with the design and development of it.

Noc2 (approved):

dsm-w3f (approved):

miketang84:
@dsm-w3f Great appreciations.

Thank you for the insights and advices.

keeganquigley (approved):
Thanks for the application @miketang84 and for answering the questions. I too, needed time to read up on this. Thanks @dsm-w3f for summing it up nicely. I had never heard of a 4-node network before so it will be interesting to see the results.

I agree with my colleague that I would rather see the app work with non-custodial keys such as a browser wallet login than Web2 OAuth methods. It would be interesting to see if you could use existing substrate identity solutions such as [Kilt](https://www.kilt.io/) or the [DID pallet](https://github.com/substrate-developer-hub/pallet-did) instead of developing your own solution.

That being said I am happy to go ahead and give you a chance. Good luck!

miketang84:
@keeganquigley Thanks.

The non-custodial way would be our prior schema when design account system.

And surely, the existing DID-like pallet would be reused once meets our demands, since that could save our time and make the code more mature and less bugs. So this is the cool part of Substrate, we can iterate based on the whole community's contributions. Cool!

PR #1428, opened 2023-01-11
Create stale_checker.yml

0xCaso:
The action checks every day at 7:00 UTC if the pull requests are on stale (>=14 days of inactivity). If yes, they get labeled, and a report is sent to the Matrix channel.

See the action repo: https://github.com/0xCaso/label-stale-pull-requests/


Noc2 (approved):
Great. Thanks!

PR #1437, opened 2023-01-13
Grant Application: Validated Streams

bojidar-bg:
#### Project Abstract

Validated Streams is a consensus primitive that aims to simplify the process of creating decentralized reactive oracles for developers of arbitrary off-chain applications.

By implementing Validated Streams on top of Substrate, we hope it could integrate with Substrate-based smart contracts, allowing developers to build more powerful and sophisticated applications that take advantage of features of both Validated Streams and Smart Contracts.

The scope of this grant application is the initial implementation of Validated Streams using Substrate.

#### Application Checklist

- [x] A BTC, Ethereum (USDC/DAI) or Polkadot/Kusama (aUSD/USDT) address for the payment of the milestones is provided in the application.

### DIFF
--- a/applications/validated-streams.md
+++ b/applications/validated-streams.md
@@ -1,0 +1,181 @@
+# Validated Streams
+
+- **Team Name:** [Comrade Coop](https://comrade.coop/)
+- **Payment Address:** [0xAE8687d444511a595Cc1EAdbFc1dFe58eB639C5f](https://etherscan.io/address/0xae8687d444511a595cc1eadbfc1dfe58eb639c5f) (DAI)
+- **[Level](https://github.com/w3f/Grants-Program/tree/master#level_slider-levels):** 1
+
+## Project Overview :page_facing_up:
+
+### Overview
+
+Validated Streams is a consensus primitive that aims to simplify the process of creating decentralized reactive oracles by providing a mechanism for verifying the authenticity and veracity of the data being used.
+
+In brief, we are providing a mechanism for proving the occurrence of events streamed from a custom application collocated with each of the validator nodes, facilitating developers of arbitrary off-chain applications in creating on-chain oracles sourced by those applications in a decentralized manner. To ensure the reliability and integrity of the data submitted on-chain, we require that at least two-thirds of the validators witness the event and confirm that it have indeed occurred off-chain.
+
+By implementing Validated Streams on top of Substrate, we hope it could integrate with Substrate-based smart contracts, allowing developers to build more powerful and sophisticated applications that take advantage of features from both Validated Streams and Smart Contracts.
+
+Validated Streams is a part of a grander vision of proactive blockchain applications that interact richly with the off-chain world while still remaining trustless and decentralized, that our team has been pushing for some years, and is now building piece-by-piece as we keep seeking the best ways to realize it.
+
+
+### Project Details
+
+![Diagram of Validated Streams, with Stream service ingesting events from an application, passing them to a gossip, which then leads to on-chain transactions, that, after block finalization, get forwarded back to the application. (validated-streams.drawio.png)](https://user-images.githubusercontent.com/5276727/211316562-ad73fdd0-0dec-4543-884e-fe60cb09ee7a.png)
+
+A Validated Streams oracle proves the occurrence of off-chain events, after they have been witnessed by at least 2/3-s of its validators.
+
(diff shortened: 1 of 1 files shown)


bojidar-bg:
Hey, thanks for giving it a look! Those are actually.. some pretty good questions!

From what I remember from internal discussions, Validated Streams is more of a communications protocol than a blockchain, and we would expect validators to be incentivized to participate through external means, not through the Validated Streams itself. As the proposal notes on line 46, though perhaps not explicitly enough, we are interested in solving the problem of having an oracle protocol, not the problem of finding data or finding validators.
However, considering that Validated Streams could work with Proof-of-Stake networks, I suspect it would be sensible to implement some kind of slashing to discourage malicious behavior, and leave the incentivization to participate to usual PoS mechanics. I'll discuss that with the team (:

A real-world example we were considering early on was organizing a DAO whose voting happens on Discord. While a DAO should generally shy away from centralized solutions, existing centralized solutions are usually a lot more comfortable for end users, hence sometimes makes sense to integrate with them. In this case, a Validated Streams oracle, likely ran by a predefined set of validators picked among the DAO's members, would allow the data stored in that centralized location to be witnessed on-chain, and from there, integrated with other on-chain parts of the DAO. The benefit (we hope) of using Validated Streams over other solutions here would be that (a) we don't depend on the centralized API directly, hence even if something were to go south there, we have an extra layer of indirection where we can stop it, and (b) we get a full, immutable history of what's happened, which gives us extra insight into operations and any problems that might arise.

And, on the note of a real world example, we've been considering dApps built entirely out of multiple Validated Streams components, where each one takes in the outputs of previous streams, calculates something, and passes on the result to the next. However, we have not fully considered the latency and security implications of doing that, so while we do want to do that in the future, we decided to scope Validated Streams to just making decentralized oracles.

Hope that helps understand more of the proposal! And, please let me know if you have more questions or things you feel should be included in the proposal document (:

Noc2 (approved):
Thanks for the quick reply and the clarification. So your solution doesn’t fully solve the oracle problem (Holy Grail ;-)). But it at least has a unique approach, and maybe the off-chain gossiping might be useful also for other projects. So I’m happy to go ahead with it.

gautamdhameja (approved):

randombishop:
Hey @bojidar-bg, thanks for this very interesting application!
- So if I am understanding this correctly: in a network with Validated Streams, the validators will run additional code to create the Oracle capability and publish data in-chain; but your protocol doesn't define an economic model to run the offchain code. Is that correct?
- If yes, is the assumption that the rewards system is not part of the protocol itself, but can be implemented amongst the validators in any ways they see fit?
- Let's say a network wants to reward oracles who pushed the correct information, would a validator be able to copy transactions from a reliable peer instead of actually doing the offchain work and free ride the work of others? Shouldn't Validated Streams implement some kind of Commit-Reveal scheme to avoid being gameable that way?
Thanks again!

Noc2:
@bojidar-bg I merged the application since it has the required approvals and is accepted, but feel free to answer @randombishop questions.

bojidar-bg:
Oh wow, merged already? :tada: :rocket: Thanks a lot, it's truly an honor to participate in w3f's grants program! :blush: We will start working on Milestone 1 right away and get back in touch on the delivery repository once that's done, then (: 

@Noc2 ~ Indeed, our solution does not cover the full extent of the oracle problem -- I'm guessing it's rather hard to design decentralized protocols where you both get useful work done and incentivize all the participants to keep going. :smile: Still, glad you see the possibility of ideas and code being used in other applications (:

@randombishop ~ That is correct, this protocol doesn't define an economic model to run the off-chain code (or even to run the on-chain code). Other people in our organization are working on a separate, orthogonal, mechanism that deals solely with establishing an economic model for running off-chain code, but they are still nailing the details down.
Hence, yes, we do expect the incentivization would be implemented amongst the validators in any ways they see fit -- potentially with a rewards system. However, I would like to note that entities can have much more nuanced motivations than one can capture with a token/price system; just like how we can observe open-source working in the human world even when it doesn't incentivize contributions financially.
For your last question, I would agree that a commit-reveal scheme could be beneficial to the protocol -- since, even from a theoretic standpoint, no one needs to know the exact extrinsics until there are 2/3 of the validators confirming them, and at that point all of them supposed to be able to reveal those on-chain. However, I would argue that a validator copying another validator's actions could also achieve that result by directly delegating its stake to the other validator, hence, I'd say it's the reward system's fault if it is gameable in that way. So, I don't think it makes sense add a commit-reveal scheme just for that purpose, at least not in the first version of the project. (:

PR #1438, opened 2023-01-13
Fix links in application templates

bojidar-bg:
Because the next person really doesn't need to have to fix URLs by hand. :sweat_smile:


Noc2 (approved):
Thanks a lot for fixing these!

PR #1449, opened 2023-01-17
Fix link in application_accepted.yml

bojidar-bg:
As accidentally spotted when clicking on links in the bot's comment. Wonder if there might be a way to quickly search the whole repo for outdated links :thinking:


Noc2 (approved):
Thanks a lot. We will look into it

PR #1479, opened 2023-01-26
Update index.md

keeganquigley:
Update index to show DatDot as grant terminated.


Noc2:
I don't think this is necessary. See elements

PR #1513, opened 2023-02-06
Fixed Case sensitive links to RFPs in Readme.md

yangwao:
Seems folder has been renamed from `rfps` to `RFPs` and I've found broken links.

Destination file 
- https://github.com/w3f/Grants-Program/blob/master/docs/RFPs/suggestion-template.md


Noc2 (approved):
Thanks for fixing this.

PR #1572, opened 2023-02-27
Move zondax maintenance application in /maintenance

0xCaso:


semuelle:
This is technically not a maintenance grant as that category didn't exist when it was signed. David set up the maintenance workflow, he would know if it's wise to move it.

Noc2 (approved):
Sounds good to me

PR #1624, opened 2023-03-20
Milestone 1 of ePassport RFP is ilegal and highly unethical, remove RFP pending rewrite

burdges:
We should update this properly.  I've need to do this anyways to actually lay out some plans.

### DIFF
--- a/docs/RFPs/Open/epassport-zk-validation.md
+++ b/docs/RFPs/Open/epassport-zk-validation.md
@@ -1,4 +1,5 @@
-# e-Passport ZK Validation
+# e-Passport ZK Validation  [ON HOLD PENDING REVISIONS]
+
 
 * **Status:** Open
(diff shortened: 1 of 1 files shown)


Noc2 (approved):
That sounds good to me. I'm pretty sure that legal would have a lot of concerns regarding this specific RFP, and I believe we were also waiting for another team to open-source their work regarding this project. So I would be fine with closing it for now.

semuelle (approved):

burdges:
I figured out that Florian and I originally discussed this before anything else existed, so even checking ePassports in any open source code looks tricky.  I think Florian wrote milestone 1 thinking of that, but I never meant for anyhting to connect to the chain there.  At some point open source non-zk epassport code materialized, making that part obsolete.

We now have open sourced zk code even.  I just need to go through their code and also see if they want to do more.  I've talked to both of them over signal and email, but not in a coherent way.

I think just marking this RFP as on-hold for now makes sense.  I'll try to figure out what makes real sense in the near future.  sorry for the mess.

PR #1636, opened 2023-03-22
Move Collator deployment RFP to Closed

semuelle:
I vote to close this RFP as we have one implementation and it's anyway rather empty.


Noc2 (approved):
Sounds good to me

PR #1657, opened 2023-04-05
Dotflow

Szegoo:
#### Project Abstract

We can most certainly agree that the future is multi-chain. As such, it is not uncommon for users to hold multiple accounts across various chains for reasons such as distinct address formats and security benefits. However, this practice presents a challenge in managing multiple addresses. Adding to the complexity is the need to verify the address of the intended recipient, as it may have changed over time.

In summary, there are two key challenges to address: the management of multiple addresses and making sure the addresses of the recipients did not change in the meantime.

To mitigate these challenges, we aim to simplify the user experience by abstracting away the complexity of address management.

We are looking to write a set of ink! contracts and develop a UI that can effectively abstract the addresses from the end user.

#### Application Checklist


Noc2 (changes requested):
Thanks a lot for the application. I have a few initial questions/comments:

- Could you add to the milestone deliverable tables that your UI will be based on the mockups? Also, could you add the programming language, etc. (like ink!), there? The milestone tables are usually the requirements of our contracts. 
- Could you remove the fact that you plan to deploy it on Astar? We usually try to be as objective as possible when it comes to different parachains and only care about the tech. 
- Just from the top of my head: What happens if you potentially run into a race condition? Especially with XCM there might be problematic edge cases. 
- It looks like people will be able to add multiple addresses per network. Is there a way to set a default payment address?

semuelle (changes requested):
Thanks for the application, @Szegoo. I have some questions and comments, for example 
- there is address book functionality in several wallets and frontends, e.g. polkadot.js.org. What's the benefit of using a separate on-chain one?
- Would I have to reconnect my wallet with a different account every time I want to send from a different address, or how could this be handled?

But overall, I have some major concerns about the idea of storing my address book for everyone else to see. The footprint we leave behind with transactions is bad enough, but allowing anyone to query the addresses I own and that I frequently interact with, across all parachains, seems rather risky. Do you know of any comparable projects, perhaps in other ecosystems, that are/were popular?

Szegoo:
@Noc2 @semuelle Thanks for reviewing my application quickly and asking me good questions.
I will try to answer all of your questions here:

> Could you add to the milestone deliverable tables that your UI will be based on the mockups? Also, could you add the programming language, etc. (like ink!), there? The milestone tables are usually the requirements of our contracts.

Yes, I will put more details into the deliverable tables.

> Could you remove the fact that you plan to deploy it on Astar? We usually try to be as objective as possible when it comes to different parachains and only care about the tech.

Sure, we are not using any Astar specific feature so our set of contracts can be deployed to any blockchain that implements the contracts pallet.

> Just from the top of my head: What happens if you potentially run into a race condition? Especially with XCM there might be problematic edge cases.

Tbh I am not sure how the race condition, in this case, would be different than in any other case if we were to create two XCM messages at the same time. Could you please elaborate more?

> It looks like people will be able to add multiple addresses per network. Is there a way to set a default payment address?

The UI mockup gives a somewhat bad example. A user can only store one address per network, so there is no need for a default payment address.

> there is address book functionality in several wallets and frontends, e.g. polkadot.js.org. What's the benefit of using a separate on-chain one?

We are aware that there are address book functionalities in several frontends but one of our main features is that inside the address book, we are not storing specific addresses, instead, we are storing `IdentityNos`. 

The reason we do this is that in case the owner of the identity that we have added to our address book changes one of his addresses the `identityNo` will still remain the same for him. So in essence we don't have to worry about updating our address book and replacing his old address with a new one. 
When sending a transaction to a specific identity the UI will query the latest address for that identity on the destination chain we specified and based on that create a transaction.

> Would I have to reconnect my wallet with a different account every time I want to send from a different address, or how could this be handled?

To send tokens to an identity the user would have to connect his wallet from which he is sending the funds, so in essence yes every time we are sending transactions from a different account we have to sign the transaction with that account. 

> I have some major concerns about the idea of storing my address book for everyone else to see. The footprint we leave behind with transactions is bad enough, but allowing anyone to query the addresses I own and that I frequently interact with, across all parachains, seems rather risky.

We are not attaching any kind of metadata(name, email, etc) to an identity so from looking at the identity on its own others could not really figure out the owner of the identity.
However, if you were to send your `identityNo` to someone that is sending you tokens frequently that person would be able to see all of your addresses.
If that is a problem we allow users to create multiple identities. So in theory this way you would be able to create an identity per every user that is sending you funds.

Szegoo:
@Noc2 @semuelle I have made the description of the milestones more detailed, hope this is better now.

Noc2 (changes requested):
Sorry for the delay here. Regarding the race condition, I meant, for example, if you update an address and at the same time, someone tries to send you money. Regarding the on-chain address book, isn’t this something that could also be easily handled by the front end? The front end could also replace the old address with the new one automatically. Also, I guess the identities you interact with most frequently are part of any block explorer, etc. But I'm generally already happy to support milestone 1.

Szegoo:
> Regarding the race condition, I meant, for example, if you update an address and at the same time, someone tries to send you money.

@Noc2 This is a very good question. Because of these types of edge cases before sending a transaction that will transfer the funds, the frontend will check whether there are any pending transactions happening that would change the address on the chain the user is transferring funds to. In case there is a pending transaction the user will get a warning and will be advised to wait for a moment and try sending the transaction again. 

> Regarding the on-chain address book, isn’t this something that could also be easily handled by the front end?

We want to store the address book information onchain. The front end will be displaying this information and will allow users to interact with the identities added to the address book.

> The front end could also replace the old address with the new one automatically. 

The address book will only be keeping the `identityNo`s, not the actual addresses. The `Identity`  contract would keep track of the user's actual addresses. The `identityNo`s will only be a reference to actual `Identities`.

> Also, I guess the identities you interact with most frequently are part of any block explorer, etc.

Block explorers deal with addresses. We want to abstract the addresses away by introducing a new structure we call `Identities`. These will hold the user's addresses and have a unique identifier called `identityNo`. The reason for this is that this way identity owners are responsible for updating their addresses in case they change. This way the senders don't need to be asking for the receiver's address once they have their `identityNo`. When having their `identityNo` they will have access to all their addresses which are kept up to date by the owner. 
Using the UI the senders will only select the `identityNo` of the receiver and specify the origin, destination chain, token, and amount they want to send. Based on this the frontend will query the actual address of the receiver and will construct a transaction.

Noc2 (changes requested):
Thanks for the quick reply. Could you add the "pending transaction" feature to the milestone deliverables? Regarding the address book, I meant a front-end that stores the identityNo as part of the UI and handles all other functionality that you describe as part of the on-chain address book. My main question here is: Is it really necessary to put this on-chain or what's the benefit of it (since on-chain storage is always more expensive and has potential privacy issues, etc.)?

Szegoo:
@Noc2 I've updated the milestone deliverables to include the "pending transaction" feature.

semuelle (commented):
Hi @Szegoo, sorry for the late response and thanks for the updates. 

> To send tokens to an identity the user would have to connect his wallet from which he is sending the funds, so in essence yes every time we are sending transactions from a different account we have to sign the transaction with that account.
> 
> We are not attaching any kind of metadata(name, email, etc) to an identity so from looking at the identity on its own others could not really figure out the owner of the identity. However, if you were to send your `identityNo` to someone that is sending you tokens frequently that person would be able to see all of your addresses.

In this case, I'm afraid I a with @Noc2 on this one. There are few upsides and at least one very big downside, namely putting all addresses that are mine or relevant to me in one place, publicly. Therefore, I won't support this application.

takahser (changes requested):
@Szegoo I understand the difficulties that come with managing multiple addresses. However, I do see the validity in the concerns raised by @semuelle regarding privacy implications. In my opinion, an on-chain address book may not be the best solution to this problem. Additionally, it's important to note that relying on wallet users to keep their addresses up-to-date could potentially lead to false security, as addresses may already be outdated.

> In this case, we can create two identities. One of them will hold my Moonbeam and Acala addresses while the other one will hold my Mangata and Interlay addresses. This way I will have two separate identities and the first project will only know about the Moonbeam/Acala identity while the other one will only know about the Mangata/Interlay identity.

I think creating multiple identities is not going to solve the problem neither, since they're still stored on-chain and anybody would be able to inspect them, incl. the addresses they contain. 

For the reasons stated above, I'm not going to approve the proposal in its current state.

randombishop:
Unfortunately, I am leaning more to @semuelle's side on this application and won't approve it.
I also think there are already existing primitives to achieve similar UX such as decentralized identity providers and things like ens pns etc.

Noc2 (changes requested):
Thanks again for the application and all the effort you put into this. However, the rest of the grants committee isn't willing to proceed with it, so I'm closing the application. Let me know if you have any questions, and feel free to apply for another grant.

Noc2:
(see above)

Szegoo:
@Noc2 We have been working on a solution to the privacy problem. I was looking to update the application. Is there a chance of re-opening this branch so that I can share the changes we have made?

Noc2:
Sure. Let me reopen it.

Szegoo:
@Noc2 As mentioned we have found a solution to the privacy problem and added a new section to the application which explains it in detail. We would really appreciate some new reviews.

Noc2 (changes requested):
Thanks for the update, and I'm personally still happy to go ahead with it. But is there a reason you have chosen AES which is a symmetric algorithm as far as I know. Doesn't it make more sense to use [Public-key cryptography](http://en.wikipedia.org/wiki/Public-key_cryptography) here?

Szegoo:
@Noc2 The reason we use symmetric key encryption is because it would be easier to manage only one key and we believe that it would be sufficient enough for this use case.

Noc2 (approved):
I personally think that for this kind of use case, public-key cryptography makes more sense. But I'm happy to approve the application anyway.

takahser:
@Szegoo regarding this:

> For the XCM transaction to work there has to be an existing XCM channel between the two blockchains.

Keep in mind that XCMP is [not rolled out yet](https://wiki.polkadot.network/docs/learn-xcm-transport#xcmp-cross-chain-message-passing).

Also, I'm a bit confused where the address book data will be stored. On one hand, you're saying it's stored locally:

> We are also going to add address book functionality, but for privacy reasons, the address book data won't be stored on the blockchain, instead, it will be stored locally on the users' device. (..)
> All of the address book data will be stored locally on the user's browser local storage.

But in other places you seem to indicate that it's stored on-chain (in the identity contract):

`Each` address will be encrypted with the associated cipher before it gets stored in the identity contract. This will ensure that no one has access to an identity's addresses simply by reading the contract's state.

takahser:
@Szegoo thanks for elaborating, that definitely helped understanding your plans better.

> Yes, we are aware of that, but we can still use HRMP.

That's fine, thanks for clearifying.

Regarding the identities, while I agree you improved security with the recent updates, I'm still not sure why you'd want them to store on-chain. Since you're locally caching them, wouldn't it be easier to create an end-to-end-encrypted communication channel between the two transacting users to transfer the identities? Can you elaborate on the reason to store them on-chain?

Szegoo:
> Regarding the identities, while I agree you improved security with the recent updates, I'm still not sure why you'd want them to store on-chain. Since you're locally caching them, wouldn't it be easier to create an end-to-end-encrypted communication channel between the two transacting users to transfer the identities? Can you elaborate on the reason to store them on-chain?

@takahser Locally in the address book we are only storing the `identityNo`s. The reason why we don't store actual addresses is actually the core idea of our project. I think this is best explained with a simple example.

**Example:**

Alice is working on a project as a contributor and she is receiving bounty rewards for her work from Bob.

Bob is usually paying her in GLMR on Moonbeam or USDT on Statemint.

Alice created an identity where she stores all of her addresses. Let's say in this example that these addresses are on Moonbeam, Statemint and Acala

She shares her `identityNo` with Bob and sends her `IdentityKey`, but only the part that gives access to her Moonbeam and Statemint address.

Bob adds Alice's identity to his address book and now he can send her tokens using the Dotflow interface.

In the case, Alice changes her Moonbeam or Statemint address she just needs to update her identity so that it now contains her new address. 
This way Bob doesn't have to worry about Alice changing her addresses since Dotflow will route all of his transactions to the proper address. So in essence our app would always query the latest address from the identity contract before sending a transaction. 

If we were to store the addresses locally as you said, Alice would always need to notify Bob about her address change and he would always need to change the addresses in his address book manually.

takahser (approved):
@Szegoo thanks for lining out the differences. I'm happy to give it a shot, good luck!

Szegoo:
> How is my identityKey sent? From what I can see, I have to copy and paste it into some communication channel I have with the receiver?

We do not get into the specifics of how the `IdentityKey` will be sent, but using some decentralized messaging apps like [Element](https://element.io/) that provide message encryption will be secure enough for sharing the key. 

> If I want to change one address but don't want those I shared the old one with to be able to read it, do I have to delete the old address and create a new one with a new cipher? Or is this not possible?

Great observation. We didn't actually provide this feature, but the solution to this would be to add an option to generate a new cipher when changing an address. We will update the application to include this option.

> Is it still true that I can only save one address per network? Why?

The reason why is because when sending a transaction we don't want the user to deal with addresses. The tokens will be routed based on the destination chain. If you take a look at the future plans we are looking to add support for that. The future plan is to add some other criteria based on which the tokens will be routed. This way it would be possible to have multiple addresses since the destination chain wouldn't be the only criteria.

> storing everything in localStorage, meaning it's transient.

After some consideration, we will bring the address book contract back. This won't affect privacy since all of the addresses are encrypted anyway. The `IdentityKey` will still be stored in the local storage and the application will expect the user to provide it when using the app from a different device for the first time. Will update the application to again include the address book contract.

> My wallet's address book automatically converts accounts between chains

Yes, but for example the Moonbeam address I use is not the one associated with my Polkadot address. So in this scenario, your address book won't be actually able to figure out my Moonbeam address.

Thanks for your comments. I will update the application to include the above mentioned changes.

Szegoo:
@semuelle As promised, I updated the application to include the changes I mentioned in my previous message. We would be really thankful if you could take a look at the changes we have made.

takahser (approved):

keeganquigley (approved):
Thanks all for the deep dives, and thanks @Szegoo for thoroughly answering all of the questions. After reading through all of the comments, I'm satisfied with your research, and it sounds like you've thought a lot about proper key management regarding AES. Perhaps the faster speed and lower computational power of symmetric encryption will benefit the app. This could be an interesting use case for storing data on-chain. Based on your experience with substrate and your reputation in the fellowship, I'm happy to approve it as well.

PR #1668, opened 2023-04-11
Amend accepted grant application Hybrid Block Explorer.

ethernomad:
This pull request is an amendment to the already accepted grant for the Hybrid Block Explorer.

During development of the first milestone it became apparent that there were some unforeseen complexities that need to be accounted for (limitations of subxt, runtime upgrades changing metadata and event types.)

The design is also more decentralized - event contents are no longer stored in the index.

Milestones 2 & 3 have been fleshed out with more features.

The timeline has been extended by 50%.


semuelle (approved):
Thanks for the update. LGTM.

PR #1171, opened 2022-09-12
Create: Kusama <> Polkadot Bridge 

Martin-Kiss:
#### Project Abstract

We are actively developing trustless bridging solutions for cross-ecosystem unification and working towards successfully bringing IBC to the Polkadot ecosystem.
Once this milestone has been realized, we will successfully establish the first trustless bridge between Kusama and Polkadot utilizing the IBC protocol.

#### Grant [level](https://github.com/w3f/Grants-Program#level_slider-levels)

- [ ] **Level 1**:  Up to $10,000, 2 approvals
- [x] **Level 2**:  Up to $30,000, 3 approvals
- [ ] **Level 3**:  Unlimited, 5 approvals (for >$100k: Web3 Foundation Council approval)

#### Application Checklist

- [x] The [application template](https://github.com/w3f/Grants-Program/blob/master/applications/application-template.md) has been copied and aptly renamed (`kusama<>polkadot-bridge.md`).
- [x] I have read the [application guidelines](https://github.com/w3f/Grants-Program/blob/master/docs/grant_guidelines_per_category.md).
- [x] A BTC, Ethereum (USDT/USDC/DAI) or Polkadot/Kusama (aUSD) address for the payment of the milestones is provided inside the application.
- [x] The software delivered for this grant will be released under an open-source license specified in the application.
- [x] The initial PR contains only one commit (squash and force-push if needed).
- [x] The grant will only be announced once the first milestone [has been accepted](https://github.com/w3f/Grant-Milestone-Delivery#process) (see the [announcement guidelines](https://github.com/w3f/Grants-Program/blob/master/docs/announcement-guidelines.md)).
- [ ] I prefer the discussion of this application to take place in a private Element/Matrix channel. My username is: `@_______:matrix.org` (change the homeserver if you use a different one)

PR #1675, opened 2023-04-17
fix: remove incorrect previous jobs

cor:
These descriptions have been submitted without our reviews and are incorrect

### DIFF
--- a/applications/CosmWasmVM-CoreProduct.md
+++ b/applications/CosmWasmVM-CoreProduct.md
@@ -105,9 +105,9 @@ We are the only team in the Substrate/Polkadot/Kusama ecosystem implementing Cos
 
 ### Team's experience
 
-- Hussein Ait Lahcen: Principal Blockchain engineer at Composable Finance, previously Core Rust developer at Parity
-- Abdullah Eryuzlu: Substrate Rust engineer at Composable Finance, previously Substrate engineer at Webb.tools
-- Cor Pruijs: Software engineer at Composable Finance, previously Software engineer at ConsenSys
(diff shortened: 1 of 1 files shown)


semuelle (approved):
Thanks for the info, @cor. No need to worry about the CLA.

keeganquigley (approved):

takahser (approved):

takahser (approved):

takahser:
I've put it on hold to double-check if this information is accurate. After that, it can be merged.

0xBrainjar2:
Hey sirs, Brainjar here, Founder & CEO of Composable Finance. We have recently undergone some internal restructuring, which has caused the delay in our submission. We value the support from the W3F and plan to update the status of the grant in the coming weeks.

Regarding the issue raised in this PR, we would like to clarify that the three team members mentioned are no longer with Composable Finance. We had planned to update this information when submitting the first milestone. Their descriptions were submitted incorrectly due to an error made by the grant writer, who used a template from a previous grant instead of starting from scratch and therefore, mistakenly included the experiences of the team members involved in the grant application submitted prior to the CosmWasm grant. (see here: https://github.com/w3f/Grants-Program/pull/1171/files#diff-9e250327ede1091f86ec3e351eeaa7b3c81260a026d5495b0fecc0fed55034cdR258)

We are sorry for any confusion this may have caused and would be happy to hop on a call to provide any further clarifications needed.

cor:
I can confirm that the information provided by @0xBrainjar2 is correct :+1:

Noc2 (approved):
Thanks a lot for the clarification here.

PR #1712, opened 2023-05-02
Proposal for IDE maintenance

lawmeskiviahs:
This is an application for the open RFP: https://github.com/w3f/Grants-Program/blob/master/docs/RFPs/Open/IDE_for_ink_Smart_Contracts.md

### DIFF
--- a/applications/maintenance/ide_maintenance.md
+++ b/applications/maintenance/ide_maintenance.md
@@ -1,0 +1,121 @@
+# IDE for ink Smart Contracts Maintenance 
+
+- **Team Name:** Antier Solutions
+- **Payment Address:** 1XhNxALoCmXF86aiwZtdDJspjWZBiRX9UfbJ4jQsWN15eSu (DOT)
+- **[Level](https://github.com/w3f/Grants-Program/tree/master#level_slider-levels):** 3
+
+## Project Overview :page_facing_up:
+
+This proposal is for a maintenance grant for the RFP - https://github.com/w3f/Grants-Program/blob/master/docs/RFPs/Open/IDE_for_ink_Smart_Contracts.md#browser-based-ide-for-ink-smart-contracts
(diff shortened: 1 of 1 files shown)


Noc2 (changes requested):
Thanks for the application. What do you mean by 2 FTE 4 PTE? 11,200 every two weeks sounds relatively expensive to me. But I will also share the application with the rest of the team.

lawmeskiviahs:
Thankyou for the reviewing the application. 2 FTE and 4 PTE means 2 full time and 4 part time. As you suggested, we have updated the costing. Kindly review the same.

lawmeskiviahs:
Thankyou for reviewing the application. 2 FTE and 4 PTE means 2 full time
and 4 part time. As you suggested, we have updated the costing. Kindly
review the same.

Noc2 (changes requested):
Thanks for the quick reply. This sounds like a lot for this project. We also discussed it internally, and it might make sense to apply for a regular grant to fix the following issues initially mostly:

- https://github.com/paritytech/ink-playground/issues/427
- https://github.com/paritytech/ink-playground/issues/197
- https://github.com/paritytech/ink-playground/issues/460
- https://github.com/paritytech/ink-playground/issues/428

Basically, making the playground compatible with different versions of ink! and automatic updates. Once this is done, we will sign a smaller maintenance agreement. We will also update the RFP accordingly.

lawmeskiviahs:
Thankyou for the update. We have updated the changes, kindly review again.

Noc2 (changes requested):
Thanks for the update and for reducing the price. But instead of submitting a maintenance grant, could you instead change it to a regular one time grant, see https://github.com/w3f/Grants-Program/blob/master/applications/application-template.md

lawmeskiviahs:
Will do.

PR #1794, opened 2023-06-07
Fix RFP images

keeganquigley:
Move img folder to avoid changing references and fix Grant Management img links.


semuelle:
We don't usually keep grant-related images in the repository. I deleted the img folder and pasted the images into the .md doc, that way they are hosted on githubusercontent.com.

PR #1860, opened 2023-07-24
Amend Polkadart

Lohann:
This amend adds a new milestone and update the value of the milestone 3, we underestimated the effort required to support signed transactions, I detailed the new challenges below:

## 1. Lack of documentation and inconsistencies
Many important components necessary for implement signed extrinsics are not documented or have ambiguous implementations, one example is Substrate BIP-39, while the original [Bitcoin BIP39](https://en.bitcoin.it/wiki/BIP_0039) is well documented, substrate's BIP39 have no documentation at all, so we must rely on the code as reference, it is also fragmented between [substrate-bip39](https://github.com/paritytech/substrate-bip39) and [Substrate](https://github.com/paritytech/substrate/blob/master/primitives/core/src/crypto.rs#L925-L951) repos, and there's a lot of complexity and tiny details there, such as derivation is algorithm dependent, the usage of [specific strings](https://github.com/paritytech/substrate/blob/polkadot-v1.0.0/primitives/core/src/ed25519.rs#L372-L376) for key derivation. Another issue is that the implementation between [polkadot-api](https://github.com/polkadot-js/common/blob/v12.3.2/packages/util-crypto/src/key/DeriveJunction.ts#L20-L35) and [subkey](https://github.com/paritytech/substrate/blob/master/primitives/core/src/crypto.rs#L185-L206) are not consistent, example:
```javascript
import { Keyring } from '@polkadot/keyring';

const keyring = new Keyring({ type: 'sr25519', ss58Format: 42 });
const account = keyring.createFromUri("//18446744073709551616");
console.log(account.address);

// Output: 5FsxbdTHtNSLE5HCLzzGVB2oEzxZB6QJuakiHSoTNWEXD1qj
```

Now using Substrate's [Subkey](https://docs.substrate.io/reference/command-line-tools/subkey/):
```shell
$ subkey inspect --scheme=Sr25519 --network=substrate "//18446744073709551616"

Secret Key URI `//18446744073709551616` is account:
  Network ID:        substrate 
 Secret seed:       0xe95232125504305f665b8e8c891e1e1e87fcbb0b4eff38a9e100862207d2ce97
  Public key (hex):  0x11ea83f6705f1e3bd3b50695227a1c64dccd5c368648f17eb80f093695028daa
  Account ID:        0x11ea83f6705f1e3bd3b50695227a1c64dccd5c368648f17eb80f093695028daa
  Public key (SS58): 5CUCLVQzLh5oxNanu3699rqaxmkiQG4kNNd4WCDfJWiqwuYk
  SS58 Address:      5CUCLVQzLh5oxNanu3699rqaxmkiQG4kNNd4WCDfJWiqwuYk
```
Both generate different addresses when the junction have more than 64bits. I believe we are the only solution in the ecosystem besides Substrate that correctly implements the Substrate-BIP39 standard:
https://github.com/rankanizer/polkadart/tree/main/packages/substrate_bip39

We found some complexity we didn't anticipate implementing other components too:
- Some signatures are non-standard [like sr25519](https://github.com/paritytech/substrate/blob/polkadot-v1.0.0/primitives/core/src/sr25519.rs#L58-L60)
- [UncheckExtrinsics uses a custom encoding/decoding format](https://github.com/paritytech/substrate/blob/polkadot-v1.0.0/primitives/runtime/src/generic/unchecked_extrinsic.rs#L296-L320)
- Signed Extensions must be manually implemented on the client-side, etc...

## 2. No Schnorrkel Library for Dart
Some cryptographic libraries like [schnorrkel](https://github.com/w3f/schnorrkel) only exists in Rust, Polkadot-api uses [native rust](https://github.com/polkadot-js/wasm/tree/v7.2.1/packages/wasm-crypto) compiled to wasm for signing transactions, we cannot do the same in dart because Flutter runs on web, mobile and desktop devices, we can't assume a native Wasm Runtime like polkadot-api does, and creating [native bindings in flutter](https://docs.flutter.dev/platform-integration/platform-channels?tab=type-mappings-c-tab) is not trivial either, it requires us to create one binding for each platform (iOS, Android, Windows, Web, MacOs, etc), and using native bindings requires Flutter developers to have the rust toolchain installed in order to use our library. Another solution is rewrite the [schnorrkel](https://github.com/w3f/schnorrkel) in pure dart, we are still deciding how to approach this issue.

## 3. Extra work for Signed Extrinsics
Substrate metadata itself doesn't provide all necessary info for building a valid transaction, RPC modules and Signed Extensions must be implemented individually for each chain [[link]](https://github.com/polkadot-js/api/tree/v10.9.1/packages/types/src/extrinsic/signedExtensions), we had to refer to different repos to understand how to properly fill up the signed extensions.

# The new Plan
While we have a lot of talented Dart/Flutter developers working in this Project, just a few of them have the necessary experience and understanding of Substrate's codebase to effectively porting it to Dart. Also I would like to highlight that Polkadart is as complex as [polkadot-api](https://github.com/polkadot-js/api) and [capi](https://docs.capi.dev/).
Below the new plan and current development status, we decide to split the milestone 3 in two parts, where on milestone 4 we will focus on [Sr25519](https://github.com/w3f/schnorrkel) and other minor functionalities.
## Milestone 3:

- [x] [Substrate BIP39](https://github.com/rankanizer/polkadart/tree/main/packages/substrate_bip39)
- [x] Ed25519
  - [x] Signatures
  - [x] Hard Junction Derivation
- [x] Ecdsa/secp256k1
  - [x] Signatures
  - [x] Hard Junction Derivation
- [ ] Substrate Signed Extensions
  - [x] CheckSpecVersion
  - [x] CheckTxVersion
  - [x] CheckGenesis
  - [ ] CheckEra (WIP)
  - [x] CheckNonce
  - [x] CheckWeigh
  - [x] ChargeTransactionPayment
- [ ] Statemint Signed Extensions
  - [ ] ChargeAssetTxPayment
- [ ] Keyring
- [ ] Transaction and events subscriptions

## Milestone 4:
- [ ] Support [Sr25519](https://github.com/w3f/schnorrkel) by rewrite it in Dart or using Native Bindings.
- [ ] Support the new Substrate's [JSON-RPC API Mega](https://forum.polkadot.network/t/new-json-rpc-api-mega-q-a/3048)
- [ ] Implement's Substrate's Multisig Accounts
- [ ] Support custom signed extensions (for non-standard extensions)
- [ ] Support custom RPC methods (for non-standard RPC modules)
- [ ] Secure Keystore:
  - [ ] [Keychain](https://developer.apple.com/documentation/security/keychain_services#//apple_ref/doc/uid/TP30000897-CH203-TP1) for iOS and MacOS
  - [ ] [Keystore](https://developer.android.com/training/articles/keystore) for Android
  - [ ] [Libsecret](https://wiki.gnome.org/Projects/Libsecret) for Linux
  - [ ] [Wincred](https://learn.microsoft.com/en-us/windows/win32/api/wincred/) for Windows
  - [ ] [WebCrypto](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto) for web

We are very proud of Polkadart, even without signed extrinsics, we already have external contributors and the project is being used in production already. Thanks to Polkadart many Flutter developers joined the Substrate ecosystem. We would love to keep supporting this project, count on us to get this done!

### DIFF
--- a/applications/Polkadart.md
+++ b/applications/Polkadart.md
@@ -184,9 +184,36 @@ Described above.
 
 ### Milestone 3 — 
 
+- **Estimated duration:** 5 months.
+- **FTE:**  2.
+- **Costs:** 20,000 USD.
+
+|Number|Deliverable|    Specification                   |
+|:---: |  ---      |             ---                    |
+| 0a.  | License          | Apache 2.0                         |
+| 0b.  | Documentation    | We will provide markdown documentation of how to install and use the API. |
+| 0c.  | Testing Guide    | We will provide sample scripts on how to use it. |
+| 0d.  | Docker           | We will provide a Dockerfile(s) that can be used to test the PoC. |
+| 0e.  | Article/Workshop | We will publish an example flutter project using the library to explain what was done and how to use it. |
+| 3a.  | Substrate Uri    | Implement Substrate's standard derivation format that applies to seeds and mnemonics. |
+| 3b.  | Encrypt/Decrypt  | Use Substrate's Bip39 password to encrypt/decrypt seeds and mnemonics. |
+| 3c.  | Ed25519          | Support Ed25519 signatures and key derivation |
+| 3d.  | Ecdsa/Secp256k1  | Support Ecdsa/Secp256k1 signatures and key derivation |
+| 3e.  | Sign & Verify    | Sign and verify messages using ed25519 and ecdsa Keypairs |
+| 3f.  | Substrate Signed Extensions | Support substrate's signed extensions |
+| 3g.  | Statemint Signed Extensions | Support statemint's signed extensions |
+| 3h.  | Transaction Subscriptions   | Support transaction subscription |
+| 3i.  | Event Subscriptions         | Support runtime event subscription |
+| 3j.  | Keyring          | Create a package that manage a set of keys in a consistent environment |
+| 3k.  | Crypto           | Utilities needed for signing, hash and encode/decode messages |
+| 3l.  | Tests            | Unit tests for deliverables above  |
+| 3m.  | Pub.dev          | Publish to pub.dev                 |
+
+### Milestone 4 — 
+
 - **Estimated duration:** 2 months.
-- **FTE:**  4.
-- **Costs:** 15,000 USD.
+- **FTE:**  2.
+- **Costs:** 20,000 USD.
 
 |Number|Deliverable|    Specification                   |
 |:---: |  ---      |             ---                    |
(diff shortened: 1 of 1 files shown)


Noc2 (changes requested):
Thanks for creating the amendment and the detailed explanation here. I'm generally happy to go ahead with it, but I have two questions/comments:
- Why did you change the FTE, and why did the cost per developer per month for the fourth milestone increase so much? 
- What's your cryptographic background/experience? I think in almost all cases it's better to use bindings for schnorrkel. Otherwise, it might also be necessary to audit the dart implementation.

semuelle:
pinging @Lohann

semuelle:
I am going to close this due to inactivity. Please let me know if I should reopen it, @Lohann.

PR #1888, opened 2023-08-04
Proposal: Storage solution on Polkadot

dtuzi:
Create proposal

#### Project Abstract

The main goal is to implement and maintain a Filecoin-like system parachain that uses DOT as the native token and is easily usable for all kinds of parachain projects in the ecosystem via XCM.
To accomplish this we are submitting this initial proposal to update our[ previous research](https://github.com/common-good-storage/report/blob/master/src/first.md) and test how XCM could be used to allow the ecosystem to use this storage network.

#### Grant [level](https://github.com/w3f/Grants-Program#level_slider-levels)

- [ ] **Level 1**:  Up to $10,000, 2 approvals
- [x] **Level 2**:  Up to $30,000, 3 approvals
- [ ] **Level 3**:  Unlimited, 5 approvals (for >$100k: Web3 Foundation Council approval)

#### Application Checklist

- [x] The [application template](https://github.com/w3f/Grants-Program/blob/master/applications/application-template.md) has been copied and aptly renamed (`project_name.md`).
- [x] I have read the [application guidelines](https://github.com/w3f/Grants-Program/blob/master/docs/Support%20Docs/grant_guidelines_per_category.md).
- [x] Payment details have been provided (bank details via email _or_ BTC, Ethereum (USDC/DAI) or Polkadot/Kusama (USDT) address in the application).
- [x] The software delivered for this grant will be released under an open-source license specified in the application.
- [x] The initial PR contains only one commit (squash and force-push if needed).
- [x] The grant will only be announced once the first milestone [has been accepted](https://github.com/w3f/Grant-Milestone-Delivery#process) (see the [announcement guidelines](https://github.com/w3f/Grants-Program/blob/master/docs/Support%20Docs/announcement-guidelines.md)).
- [ ] I prefer the discussion of this application to take place in a private Element/Matrix channel. My username is: `@_______:matrix.org` (change the homeserver if you use a different one)

PR #1889, opened 2023-08-04
xNFT updates

anshulWeb3:
This pull request is to amend the deliverables for xNFT pallet grant. We have replaced two functions namely transferSingleWithFee and transferMultiWithFee functions with transferCollection and transferCollectionOwnership functions. 
The transferCollection function will be used to transfer collection from one chain to another.
While the transferCollectionOwnership will transfer the ownership of a collection across chains.
The reason behind this being: the instructions of xcm which we are using to transfer nfts is transact  so for now we don't have any requirements related to fees. 
Some instructions of xcm like buy execution requires fee parameter, but we aren't using any such instructions in our pallet.
Also, added names of two team members.

### DIFF
--- a/applications/xNFT.md
+++ b/applications/xNFT.md
@@ -1,7 +1,7 @@
 # xNFT
 
 - **Team Name:**  Starks
-- **Payment Address:**  0x2492237FA7698B8F3B35F2be4be3B1128439Ec8d USDT
+- **Payment Address:** 0x2492237FA7698B8F3B35F2be4be3B1128439Ec8d (USDC) **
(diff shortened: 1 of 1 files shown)


Noc2 (changes requested):
Thanks for your interest in the grants program. But this doesn't look like an amendment to me, but rather like a new grant application, correct? In this case, could you also fill out this template: https://github.com/w3f/Grants-Program/pull/1888#issue-1836561097 as part of the PR? If it's actually an amendment, could you instead update the original application and not add a new file to the repo?

PR #1893, opened 2023-08-07
xNFT updates

anshulWeb3:
This pull request is to amend the deliverables for xNFT pallet grant. We have replaced two functions namely transferSingleWithFee and transferMultiWithFee functions with transferCollection and transferCollectionOwnership functions. 
The transferCollection function will be used to transfer collection from one chain to another.
While the transferCollectionOwnership will transfer the ownership of a collection across chains.
The reason behind this being: the instructions of xcm which we are using to transfer nfts is transact  so for now we don't have any requirements related to fees. 
Some instructions of xcm like buy execution requires fee parameter, but we aren't using any such instructions in our pallet.
Also, added names of two team members.

### DIFF
--- a/applications/xNFT.md
+++ b/applications/xNFT.md
@@ -18,27 +18,36 @@ The Crosschain NFT Pallet is a unique initiative designed to facilitate the smoo
 
 #### The xNFT Pallet provides the following extrinsics (functions):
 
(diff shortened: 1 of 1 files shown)


Noc2 (approved):
Thanks for the update. I will share it with the rest of the team.

semuelle (approved):

PR #1924, opened 2023-08-22
Update RFPs

keeganquigley:
Updated RFP statuses and location.


Noc2 (approved):
Great! Thanks a lot!

semuelle:
We should consider not moving the files for each status update. That way, the links stay valid.

PR #2031, opened 2023-10-04
Update index.md - Waves 9 - 12 status review

dsm-w3f:


semuelle (changes requested):
Hey Diogo. We only use the Terminated column here to indicate when a project was terminated due to breach of contract or similar offenses, not for simple cancellations. Could you review those again. Sorry if I never mentioned this!

semuelle:
That's a lot of cancellations :laughing:. Thanks!

PR #2102, opened 2023-11-10
Liisa Portfolio Tracker - Updated Deadlines, Chain suport and Cloud provider

LiisaNFT:
Dear all, 

After consulting with the W3F team, we hereby submit the following changes to our initial application:

**1. Removing Moonsama and Unique Network support:**

- With Moonsama, we're having trouble identifying sales and values due to complexities in decoding the Raresama marketplace's functions. Our talks with the Moonsama/Raresama team are ongoing, but we haven't found a fix yet. 

- For Unique Network, the lack of an archive on Subsquid means we can't index data for Unique NFTs right now. We're in active talks with the teams at Subsquid and Unique Marketplace to sort this out, but there's no solution yet.

- The remaining chains/protocols have been integrated successfully. We think focusing on what we can deliver right now for operational chains is better so that we can resume to the next stage of the project.

- We have updated the Components section accordingly, as well as the milestone tables.

**2. Increasing the project’s timeline** 

- In light of these issues, we're proposing an adjusted timeline. As suggested by the W3F team, we’re adding one month to Milestone 1, due to the challenges we have faced, and an additional month to Milestone 2, to account for unforeseen challenges that might arise in this next stage.

- We have updated the Development roadmap / milestone tables accordingly

**3. Migration to GCP**

- We will no longer be using AWS, but GCP to host the solution.

- We have updated the Components section accordingly.


semuelle (changes requested):
Thanks for the submission, @LiisaNFT, but could you submit a pull request from a fresh fork? Since you are trying to add a document that already exists, it's conflicting. Also, it's not possible to see what changes are actually made to the document.

LiisaNFT:
Thank you @semuelle , we've just submited a new PR from a fresh fork

PR #2103, opened 2023-11-11
Liisa portfolio tracker application update

LiisaNFT:
Dear all,

After consulting with the W3F team, we hereby submit the following changes to our initial application:

1. Removing Moonsama and Unique Network support:

- With Moonsama, we're having trouble identifying sales and values due to complexities in decoding the Raresama marketplace's functions. Our talks with the Moonsama/Raresama team are ongoing, but we haven't found a fix yet.

- For Unique Network, the lack of an archive on Subsquid means we can't index data for Unique NFTs right now. We're in active talks with the teams at Subsquid and Unique Marketplace to sort this out, but there's no solution yet.

- The remaining chains/protocols have been integrated successfully. We think focusing on what we can deliver right now for operational chains is better so that we can resume to the next stage of the project.

- We have updated the Components section accordingly, as well as the milestone tables.

2. Increasing the project’s timeline

- In light of these issues, we're proposing an adjusted timeline. As suggested by the W3F team, we’re adding one month to Milestone 1, due to the challenges we have faced, and an additional month to Milestone 2, to account for unforeseen challenges that might arise in this next stage.

- We have updated the Development roadmap / milestone tables accordingly

3. Migration to GCP

- We will no longer be using AWS, but GCP to host the solution.

- We have updated the Components section accordingly.

### DIFF
--- a/applications/LiisaPortfolioTracker.md
+++ b/applications/LiisaPortfolioTracker.md
@@ -5,7 +5,7 @@
 - **[Level](https://github.com/w3f/Grants-Program/tree/master#level_slider-levels):** 2
 
 
-## Project Overview :page_facing_up:
+## Project Overview :page_facing_up: 
 
 Liisa operates as a multi-chain NFT data analytics platform, engineered to empower collectors and investors in making proficient, data-driven decisions. 
 
@@ -20,7 +20,7 @@ The portfolio tracker is designed to accommodate the diverse needs of a broad us
 
 ### Architecture
 
-<img width="960" alt="PortoflioTrackerArchitectureNew" src="https://github.com/LiisaNFT/W3F-Grants-Program/assets/139144686/4990a0db-ea37-4458-b30f-b10b6c464d98"/>
+![Architecture New](https://github.com/LiisaNFT/W3F-Grants-Program/assets/139144686/5ded60b3-0409-44fb-a619-b47c0c6e7de7)
 
 
 ### Technologies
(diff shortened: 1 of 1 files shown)


keeganquigley:
Thanks for the updates @LiisaNFT looks good to me. I will mark the amendment as ready for review and ping the committee.

keeganquigley (approved):

Noc2 (approved):

takahser:
@LiisaNFT thanks for the update. I'm fine with most changes, but I didn't quite understand the issue with Moonsama. For example, it's [compatible with subsquid](https://docs.subsquid.io/evm-indexing/supported-networks/), so you'd get the sale data from the transactions, right? Also, what did your contacts at Moonsama say, if you don't mind asking?

image

LiisaNFT:
Dear @takahser, thank you for your feedback.

In regards to Moonsama, Subsquid is indeed capable of indexing the chain and we are currently able to index ERC721 and ERC1155 transfers, since the ABIs for the token standards are equal across all EVM chains, meaning we can "listen" to and decode these transactions. 

Our challenge comes from identifying the ABIs from the marketplaces (Raresama in this case) so that we're able to decode transactions on these marketplaces, such as sales - which include not only the transfer of the ERC721/1155 token, but other information such as the sales price (a native token transfer), marketplace fees, etc. 

The Moonsama team has told us that they are not able to provide us with their marketplace's ABI. They also mentioned they are developing (or planning to develop) their own API for this purpose, which isn't yet available.

Hope this answers your question, let us know if you need any further clarification!

takahser (approved):
@LiisaNFT thanks for your reply.

> The Moonsama team has told us that they are not able to provide us with their marketplace's ABI. 

If that's the case, I'm happy to approve the amendment.

PR #2129, opened 2023-11-28
Update lastic-price-simulation-2.md to add additional contributor.

poppyseedDev:
Added Pierina Ponce.


Noc2 (approved):
Thanks for the update. I will share it with the rest of the day.

PR #2130, opened 2023-11-28
Integrate Sumsub

semuelle:


Noc2 (dismissed):
Looks good. Should we also add to the template and/or guidelines? Just to make sure people are aware of it.

semuelle:
> Looks good. Should we also add to the template and/or guidelines? Just to make sure people are aware of it.

Will do, good point.

PR #2229, opened 2024-02-22
Update address in Ink Playground IDE improvement application

radhe-zeeve:
Updated office address to match address in latest invoices.


Noc2 (approved):
Thanks for the update. I will share it with the rest of the team.

takahser (approved):

PR #2248, opened 2024-03-14
confirm vested DOT

PieWol:


semuelle:
Good idea, but I would put it into the [pull request template](https://github.com/w3f/Grants-Program/blob/master/.github/pull_request_template.md). It already has a checklist.

PR #2262, opened 2024-03-26
Update faq.md

semuelle:
Is this understandable? Suggestions welcome.


docs/faq.md
```diff
@@ -17,7 +17,7 @@ title: 🙋 FAQ
...
+  Every milestone payment will have two parts: 70% in the currency of your choice (see the application template for options) and the remaining 30% in DOT, vesting linearly over two years. For this, we need a Polkadot address in the grant agreement and, when it comes to milestone payments, two invoices: one for the currency part, one for the DOT part. Since the grant agreements are denominated in USD, we ask for invoices denominated in USD; we will convert the 30% to DOT at the day of payment based on the previous day's closing price on [CoinMarketCap](https://coinmarketcap.com). Our [invoice form](https://docs.google.com/forms/d/18BdmYEt8gvcC-5f7TMHdJftBr9dEeA2_lUfIqkH7it8/viewform) will walk you through the process.
```
PieWol, line 20:
```suggestion
  Every milestone payment will have two parts: 70% in the currency of your choice (see the application template for options) and the remaining 30% in DOT, vesting linearly over two years. For this, we need a Polkadot address in the grant agreement. It has to be a native Polkadot address as this is is the only way to receive vested DOT that can be staked. This wouldn't be possible with DOT paid on AssetHub. Upon milestone payments please submit two invoices: one for the currency part, one for the DOT part. Since the grant agreements are denominated in USD, we ask for invoices denominated in USD; we will convert the 30% to DOT at the day of payment based on the previous day's closing price on [CoinMarketCap](https://coinmarketcap.com). Our [invoice form](https://docs.google.com/forms/d/18BdmYEt8gvcC-5f7TMHdJftBr9dEeA2_lUfIqkH7it8/viewform) will walk you through the process.
```

PieWol: something like this?

Noc2: Looks good to me

Noc2 (dismissed):
I would integrate @PieWol changes. Otherwise, looks good.

docs/faq.md
takahser, line 20: I think we also support CHF, EUR (not just USD). So "denominated in FIAT" would be more accurate I think.

semuelle: you're right

semuelle: actually, [no](https://github.com/w3f/Grants-Program/blob/03ccfc8c06ab45671701afc0d3a2ebf2bdbbdb0b/applications/application-template.md?plain=1#L126). We support payment in those, but applications should be denominated in USD.

PR #2274, opened 2024-04-03
Update substrate-tutorials.md

eltociear:
usefull -> useful


semuelle:
Thanks for the PR, @eltociear, but it's not actually necessary to spell check old grant agreements.

PR #2299, opened 2024-05-05
Add Tarek to Solang Playground contributers

salaheldinsoliman:
@tareknaser Has been a involved with the Solang ecosystem, in a Hyperledger mentorship, and will be a main contributer to Solang Playground.


salaheldinsoliman:
I have read and hereby sign the Contributor License Agreement.

semuelle (approved):

applications/Solang_Playground.md
```diff
@@ -1,8 +1,10 @@
 # Solang Playground
 
 
-- **Team Name:** Salaheldin Soliman
-- **Payment Address:** 0x9036b20920ddf473e796d24afd0fe04d4dde0b68
+- **Team Name:** Salaheldin Soliman and Tarek Elsayed
+- **Payment Address:**
```
PieWol, line 5: You added another payment address. Does this mean you both want to receive 50% of the total amount?

PieWol (changes requested):

Noc2 (approved):

applications/Solang_Playground.md
salaheldinsoliman: Nope, but after each milestone, we will file two invoices (one per each contributer) based on our agreement.
If the exact split is to be amended to this file that should be better of course

PieWol: Got it, thanks for clarifying.

PR #2302, opened 2024-05-07
 Update bubble plugins list

paulogustavopeixoto:
Hello guys,
After reviewing the wallets documentation we identified 2 wallets that we won't be able to integrate: Parity signer - which we found out they do only native wallets , and Polkadot-js which already exists. Let us know if you have any questions regarding this.


paulogustavopeixoto:
Hello @semuelle . I am getting stuck in the CLA verification. I sign with my verification, but it seems it does not want to accept liisa's. Liisa used to be an account but was converted into an organization. I added the email to my account but still no success. Do you have any idea on how to fix this?

semuelle (approved):
Thanks for the amendment. Don't worry about the GitHub action, it's just for safety but not actually necessary.

takahser (approved):

Noc2 (approved):

PR #2309, opened 2024-05-22
Amend grant agreement secret management

Ra33it0:
# Amendment
Upon further evaluation, it turned out that requiring Vault to run a validator adds unnecessary complexity for no real benefit to users that don't already utilize Vault.  Effectively, at the moment, Vault would only be used to deliver already existing credentials to a single validator.  So even though it would be possible to automatically spin up a Vault cluster alongside the validator, it would neither increase security nor ergonomics of Dotnix.  The fact that the validator secrets have to exist unencrypted on the filesystem anyway allows for a simpler implementation where the user can manage secrets using a command line utility with no need for additionally running services.

This does not prevent a future integration with Vault, but creates a base line and a convenient interface for the user.

We already have implemented a succeeding test scenario that provides secrets from Vault to the validator.
This work has been pushed to will be kept for future reference in a separate branch: https://github.com/sporyon/dotnix-core/compare/vault


semuelle (changes requested):
Thanks for submitting the amendment, @Ra33it0. Could you remove the application template content from the pull request body and instead explain what changes this amendment makes and why you want to do them?

semuelle (approved):

PieWol (approved):
Thanks for taking your time and adjusting the application!

Noc2 (approved):

PR #2376, opened 2024-08-22
Amendment for Subcoin milestone3

liuchengxu:
In light of recent developments, it has become evident that fully syncing to the tip of the Bitcoin network and enabling new nodes to perform fast sync to the latest Bitcoin state is more challenging than initially anticipated, caused by the huge state of UTXO set (over 12GiB, much larger than any existing Substrate chains to my knowledge). We'll also discuss the fast sync challenge in the article delivery at this milestone. As a result, I would like to propose adjusting the delivery goal for this milestone.

The most significant known blocker is https://github.com/paritytech/polkadot-sdk/issues/4. Other underlying issues may also contribute to the difficulty. Recent experiments have shown that fast sync from around block height 580,000 is currently infeasible, succeeding only on machines with 128GiB of memory (https://github.com/paritytech/polkadot-sdk/issues/5053#issuecomment-2296043492), which is impractical for most users. Nevertheless, we have successfully demonstrated that decentralized fast sync is possible within a prototype implementation.

While syncing to the Bitcoin network's tip remains a future target, addressing the existing technical challenges will require substantial R&D efforts. We remain committed to exploring potential solutions, including architectural changes and contributing to resolving issue https://github.com/paritytech/polkadot-sdk/issues/4,

### DIFF
--- a/applications/Subcoin.md
+++ b/applications/Subcoin.md
@@ -129,7 +129,7 @@ This project is planned as 3 milestones, completed in three months.
 | 0c. | Testing and Testing Guide | Core functions will be fully covered by comprehensive unit tests to ensure functionality and robustness. In the guide, we will describe how to run these tests. |
 | 0d. | Docker | We will provide a Dockerfile(s) that can be used to test all the functionality delivered with this milestone. |
 | 0e. |	Article | We will publish an article that explains what was done/achieved as part of the grant. |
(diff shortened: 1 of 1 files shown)


PieWol:
Hey @liuchengxu ,
thanks for reaching out about this issue. Simply out of interest, have you tried to allow for a huge swap file so that this might be successful even on machines without 128gb of RAM?

Noc2 (approved):
Happy to go ahead with this.

PieWol (approved):
Thank you for swiftly conducting the experiment and laying out the results. The change to the grant agreement sounds very reasonable to me.

keeganquigley (approved):

PR #2380, opened 2024-09-01
New funding flowchart

PieWol:
An updated funding flowchart inspired by [Shawn Tabrizi](https://forum.polkadot.network/t/is-the-treasury-polkadots-biggest-vc/9015)

### DIFF
--- a/README.md
+++ b/README.md
@@ -260,19 +260,19 @@ The following flowchart gives a rough, oversimplified view of how the W3F Grants
 ```mermaid
 flowchart LR
     A(Project Focus)
-    A -->|Development| B[Stage of Development]
-    A -->|Research| C[Grants Program]
-    A -->|Other| D[Treasury]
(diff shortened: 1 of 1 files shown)


Noc2 (changes requested):
What do you mean by immediate public value?

PieWol:
> What do you mean by immediate public value?

By that I mean anything which results in an immediate benefit for the Polkadot ecosystem. I‘d say this is what it boils down to when the treasury and their voters have to evaluate what they are funding. 

The term is taken from this [forum post](https://forum.polkadot.network/t/is-the-treasury-polkadots-biggest-vc/9015) by Shawn Tabrizi.

Noc2 (changes requested):
It's too unspecific. At least, this would not help me understand, if I should apply for a grant or apply via the treasury.

PieWol:
postponed

PR #2417, opened 2024-10-08
fix rfp status in file

PieWol:
After only modifying the RFP overview section I realized that the status also needs to be changed within the RFP .md file itself.


semuelle (approved):
Thanks

PR #2450, opened 2024-11-16
terminate java substrate client

PieWol:
Hey @keeganquigley ,
the update to the index to display the grant termination. see #2443


keeganquigley:
Thanks @PieWol although afaik typically we only update this if it was terminated due to a breach of terms/conditions and not due to inactivity.

PR #2453, opened 2024-11-18
Change duration of the last milestone

dastansam:


keeganquigley (changes requested):
Thanks @dastansam could you please explain the reason for this change and give an update on the current progress?

dastansam:
> Thanks @dastansam could you please explain the reason for this change and give an update on the current progress?

the main reason is availability and an additional effort that was spent to upgrade some on-chain and offchain components. And current progress is that, we are done with the auxillary changes and working on completing the UI component

semuelle (approved):

keeganquigley (approved):

Noc2 (approved):

PR #2455, opened 2024-11-21
Add tips for good applications

semuelle:
Please review and add what you can think of.

### DIFF
--- a/docs/help.md
+++ b/docs/help.md
@@ -3,6 +3,17 @@ sidebar_position: 7
 title: 💡 Help
 ---
 
+### Tips for submitting a good grant application
+
+- Instead of just checking off each item on the application form, think about what information a reader might need to understand what you are planning to build _and why_. Speaking from experience, the most common questions the committee asks are 
+  - whether feedback about the project has been collected from the community, target audience and/or potential investors,
+  - what the status of the project is going into the grant and what its status will be afterwards (i.e. what your plans beyond the grant are), and
+  - what your motivation to build _and maintain_ the project is.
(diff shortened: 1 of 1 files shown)


docs/help.md
```diff
@@ -3,6 +3,17 @@ sidebar_position: 7
...
+  - what your motivation to build _and maintain_ the project is.
+- Take the time to read our **supplementary material**, such as our [guidelines](https://grants.web3.foundation/docs/Introduction/intro), [FAQs](https://grants.web3.foundation/docs/faq) and [guidelines for popular grant categories](https://grants.web3.foundation/docs/Support%20Docs/grant_guidelines_per_category) to understand what we're looking for.
+- Be liberal in the description of your broader vision, but be specific about how you plan to achieve your goals and which parts will be covered by the grant.
+- The more information you can provide in the application, the better. The grant application "only" needs to contain all information required to define the grant, but we are happy to also take supporting documents into account during the evaluation, such as whitepapers, academic papers or presentations.
+- Collect feedback from your target audience. The more you can show that you made an effort to confirm market fit and verification of your approach (do people in the ecosystem agree? How is this problem solved in other ecosystems?), the easier our job.
```
keeganquigley, line 15:
```suggestion
- Collect feedback from your target audience. The more you can demonstrate that you made an effort to confirm market fit and verification of your approach (do people in the ecosystem agree? How is this problem solved in other ecosystems?), the easier our job becomes.
```

PieWol:
Great content. The only downside I see is that our docs are already quite a lot of text. I think the overlap with https://grants.web3.foundation/docs/Introduction/intro is too much for this to deserve it's own additional entry. How about we extend the guideline content instead? We could also split the current guidelines to only feature formal necessities and move all current tips from there into this newly proposed ```Tips for submitting a good grant application``` section. Let me know what you think :)

semuelle:
I was just going to propose moving the "your project will have better chances to be accepted if" section here, but you beat me to it. It's poorly structured anyway: "This is what we want" - "This makes it better" - "This is also required". 

I'll give it a try.

